PR #1, opened 2023-12-11
initial

janishorsts:


lex.go
```diff
@@ -0,0 +1,653 @@
...
+			return l.emitItem(mk(itemReserved, s))
+		case len(s) == 0 && isReservedFirstChar(v), isReservedChar(v):
+			s += string(v)
+		}
+	}
+}
+
+// helpers
+
+// isAlpha returns true if r is alpha character.
```
klavinsjanis, line 526: alpha or alphabetic character?

lex_test.go
```diff
@@ -0,0 +1,314 @@
...
+				mk(itemText, "Submit"),
+				tokenExpressionOpen,
+				mk(itemClosingFunction, "button"),
+				tokenExpressionClose,
+				tokenEOF,
+			},
+		},
+		{
+			name:  "variable",
+			input: "{$count :math:round}",
```
klavinsjanis, line 83: is ':' character permitted in the standalone function name?

janishorsts: yes, MF2 support namespaces in identifiers now - https://github.com/unicode-org/message-format-wg/blob/main/spec/message.abnf#L86

lex_test.go
```diff
@@ -0,0 +1,332 @@
...
+				mk(itemExpressionClose, "}"),
+				tokenEOF,
+			},
+		},
+		{
+			name:  "quoted literal",
+			input: "{|Hello, world!| :uppercase}",
+			expected: []item{
+				tokenExpressionOpen,
+				mk(itemLiteral, "Hello, world!"),
```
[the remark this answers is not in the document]
janishorsts: Not sure if differentiation is required. What are the benefits?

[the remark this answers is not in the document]
klavinsjanis: Not losing information when marshaling from MF2 to AST and back.

lex_test.go
```diff
@@ -0,0 +1,332 @@
...
+				mk(itemWhitespace, " "),
+				mk(itemVariable, "hostName"),
+				mk(itemWhitespace, " "),
+				mk(itemOperator, "="),
+				mk(itemWhitespace, " "),
+				mk(itemExpressionOpen, "{"),
+				mk(itemVariable, "host"),
+				mk(itemExpressionClose, "}"),
+				mk(itemWhitespace, " "),
+				// .local $h = {$hostName}
```
klavinsjanis, line 207: shouldn't this be: .local $h = {$host} ?

lex_test.go
```diff
@@ -0,0 +1,332 @@
...
+				mk(itemVariable, "host"),
+				mk(itemExpressionClose, "}"),
+				tokenEOF,
+			},
+		},
+		{
+			name:  "input declaration",
+			input: ".input {$host} .input {$user}",
+			expected: []item{
+				// .input {$hostName}
```
klavinsjanis, line 224: // .input {$host}

lex_test.go
```diff
@@ -0,0 +1,332 @@
...
+				mk(itemKeyword, "input"),
+				mk(itemWhitespace, " "),
+				mk(itemExpressionOpen, "{"),
+				mk(itemVariable, "host"),
+				mk(itemExpressionClose, "}"),
+				mk(itemWhitespace, " "),
+				// .input ${host}
```
klavinsjanis, line 231: // .input ${user}

lex_test.go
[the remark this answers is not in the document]
janishorsts: The only lost information is about how it was represented in the original syntax.

When encoding back to MF2, you still have to decide to use unquoted or quoted - `Hello` or `|Hello|`.

E.g. you have unquoted `Hello` in the original. Later, it was changed to `Hello, world!` in the AST. You cannot use unquoted anymore. You have to use quoted literal now - `|Hello, world!|`.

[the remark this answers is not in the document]
klavinsjanis: It would be nice to keep the original syntax intact for testing when no modifications are made.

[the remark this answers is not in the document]
janishorsts: What scenarious are you talking about, @klavinsjanis?

It would make it harder for AST consumer. It enforces decision making onto the consumer when making changes to AST - to use quoted or unquoted.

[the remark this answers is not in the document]
klavinsjanis: ok, that makes sense. I was just wondering if we will get the same result after unmarshal & marshal back to MF2.
MF2 "{hello :uppercase}" -> AST ->  MF2  "{hello :uppercase}"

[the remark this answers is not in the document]
janishorsts: For the sake of simplicity, we might use quoted literals when marshalling. We'll see.

lex_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+				items = append(items, v)
+				if v.typ == itemEOF || v.typ == itemError {
+					break
+				}
+			}
+
+			// assert
+
+			assert.Equal(t, test.expected, items)
```
VladislavsPerkanuks, line 313: how about, instead of checking slice versus slice, we check every element individually
`expected[i] == items[i]`

and if it fails then we call `lognext()`

Reason: decrease test output clutter

janishorsts: Can you please elaborate on the clutter?

Unless you specify verbose output for tests you don't see it. Only when it fails.

VladislavsPerkanuks: Yeah I see, but I thought we were running tests with `verbose` flag as the default, which seems not to be the case here.

Running without verbose flags, indeed logs this only when test fails.

lex.go
```diff
@@ -0,0 +1,653 @@
...
+
+// Keywords.
+const (
+	keywordMatch = "match"
+	keywordLocal = "local"
+	keywordInput = "input"
+)
+
+// item is a token returned by the lexer.
+type item struct {
```
klavinsjanis, line 83: what benefits of naming this type -  item instead of token?

lex.go
```diff
@@ -0,0 +1,653 @@
...
+func mk(typ itemType, val string) item {
+	return item{typ: typ, val: val}
+}
+
+// lex creates a new lexer for the given input string.
+func lex(input string) *lexer { return &lexer{input: input} }
+
+// lexer is a lexical analyzer for MessageFormat2.
+//
+// See https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md
```
klavinsjanis, line 98: Since we can't follow any versions, perhaps we could add information about the date when syntax was used to build the implementation.

janishorsts: This is something I'm trying to figure out. It's not gonna be the date as it is not deterministic.

Either git commit value or copy of an ABNF file in this repo.

VladislavsPerkanuks: just link to abnf from commit this was made from, should be good
e.g.

https://github.com/unicode-org/message-format-wg/blob/7c00820a0462679eba696181c45bfadb43d2eedd/spec/message.abnf

klavinsjanis (approved):

lex_test.go
```diff
@@ -0,0 +1,335 @@
...
+
+func Test_lex(t *testing.T) {
+	t.Parallel()
+
+	for _, test := range []struct {
+		name     string
+		input    string // MessageFormat2 formatted string
+		expected []item
+	}{
+		{
```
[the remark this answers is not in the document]
janishorsts: Yes, I anticipate more changes to MF2 ABNF and I'd like to unblock us as quickly as possible.

I may add unhappy tests later, in another PR.

VladislavsPerkanuks (approved):

PR #5, opened 2023-12-14
Implement Parser

VladislavsPerkanuks:
~**Waiting on #8, before be able to merge**~

# What is Done
- Interfaces and Implementations of nodes (almost all of them)
- Parsing of simple message and complex message
- Tests that cover 99% of Translate project workflow

# What is **not** Done
- Some nodes are missing: i.e `ReservedStatement, PrivateUseAnnotation, ReservedAnnotation`
- Their implementation
- **Error handling**: for now it just skips, tokens until it meets the right one. Basically, only a happy path works.
- ~Lexer `nextItem()`, goes into infinite loop, when MF2 contains multiline e.g.~ **My bad I tested with newline before `.match...` which is not correct syntax**
- ~Lexer `nextItem()` goes into an infinite loop, when a complex message contains only `QuotedPattern` e.g `{{Hello, { |literal| } World!}}`.~ **Fixed in #7**
- Only functions with prefix `:` are implemented.


parse.go
```diff
@@ -0,0 +1,395 @@
...
+		if item.typ == itemEOF {
+			return nil
+		}
+
+	}
+
+	return errors.New("too many tokens. infinite loop ?")
+}
+
+func new(lexer *lexer) *parser {
```
janishorsts, line 49: We should not override system-predefined identifiers. This configures `predefined` lint rule https://github.com/expect-digital/go-mf2/pull/6

parse.go
```diff
@@ -0,0 +1,395 @@
...
+	var message Message
+
+	// TODO: parse error handling
+	if isFirstKeyword || isFirstQuotedPattern {
+		message = p.parseComplexMessage()
+	} else {
+		message = p.parseSimpleMessage()
+	}
+
+	return message, nil
```
janishorsts, line 76:
```suggestion
	// TODO: parse error handling

	// Determine if the input is a complex message.
	if p.items[0].typ == itemKeyword || p.items[0].typ == itemQuotedPatternOpen {
		return p.parseComplexMessage(), nil
	}

	return p.parseSimpleMessage(), nil
```

parse.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	return Identifier{Namespace: ns, Name: name}
+}
+
+// helpers
+
+// strAsNumber tries to parse a string as a number.
+// Returns int64 or float64 if successful, otherwise returns nil.
+func strAsNumber(s string) interface{} {
```
janishorsts, line 385: Number literal has the following definition. 

```abnf
; number-literal matches JSON number
; https://www.rfc-editor.org/rfc/rfc8259#section-6
number-literal = ["-"] (0 / ([1-9] *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]
```

`strconv.ParseInt` does not parse according to the given RFC (it can parse `inf` and other values that are not part of the RFC). Please use `json.Unmarshal()` instead.

Always unmarshal to `float64`

ast.go
```diff
@@ -0,0 +1,274 @@
...
+
+type PlaceholderPattern struct {
+	Pattern
+
+	Expression Expression // LiteralExpression, VariableExpression, or AnnotationExpression
+}
+
+// --------------------------------Expression----------------------------------
+
+type LiteralExpression struct {
```
janishorsts, line 108: Can we simplify and have only one `Expression` struct? Expression is annotation with optional (variable | literal).

janishorsts:
```
.match { $variable :number }

1 {{Hello { $variable} world}}
* {{Hello { $variable } worlds}}
```

Works for me.

```go
		{
			input: ".match { $variable :number }\n\n1 {{Hello { $variable} world}}\n* {{Hello { $variable } worlds}}",
			expected: []item{
				mk(itemKeyword, ".match"),
				mk(itemWhitespace, " "),
				mk(itemExpressionOpen, "{"),
				mk(itemWhitespace, " "),
				mk(itemVariable, "$variable"),
				mk(itemWhitespace, " "),
				mk(itemFunction, ":number"),
				mk(itemWhitespace, " "),
				mk(itemExpressionClose, "}"),
				mk(itemWhitespace, "\n\n"),
				mk(itemLiteral, "1"),
				mk(itemWhitespace, " "),
				mk(itemQuotedPatternOpen, "{{"),
				mk(itemText, "Hello "),
				mk(itemExpressionOpen, "{"),
				mk(itemWhitespace, " "),
				mk(itemVariable, "$variable"),
				mk(itemExpressionClose, "}"),
				mk(itemText, " world"),
				mk(itemQuotedPatternClose, "}}"),
				mk(itemWhitespace, "\n"),
				mk(itemLiteral, "*"),
				mk(itemWhitespace, " "),
				mk(itemQuotedPatternOpen, "{{"),
				mk(itemText, "Hello "),
				mk(itemExpressionOpen, "{"),
				mk(itemWhitespace, " "),
				mk(itemVariable, "$variable"),
				mk(itemWhitespace, " "),
				mk(itemExpressionClose, "}"),
				mk(itemText, " worlds"),
				mk(itemQuotedPatternClose, "}}"),
				mk(itemEOF, ""),
			},
		},

```

ast.go
VladislavsPerkanuks: We could, but I would rather not.
**Reason**: I want our `AST` tree to look like `ABNF` syntax as much as possible, and combining in one expression compromises that.

parse.go
```diff
@@ -0,0 +1,419 @@
...
+func (p *parser) parseIdentifier() Identifier {
+	full := strings.Split(p.current().val, ":")
+
+	var (
+		ns   string
+		name string
+	)
+
+	//nolint:gomnd
+	// TODO: error handling: case unexpected length
```
janishorsts, line 403: The lexer will error on the invalid function name, you don't have to worry about it.

janishorsts (approved):

PR #13, opened 2023-12-18
AST string representation

VladislavsPerkanuks:
AST to MF2 string representation

~**Waiting for #5**~

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -310,7 +310,7 @@ type ReservedStatement struct {
 	// TODO: Implementation
 }
 
-func (ReservedStatement) String() string { return "TODO" }
+func (ReservedStatement) String() string { return ".RESERVED NOT IMPLEMENTED { TODO }" } // TODO: Implement
 
 // --------------------------------VariantKey----------------------------------
 
--- a/ast.go
+++ b/ast.go
@@ -411,10 +411,14 @@ const (
 
 // sliceToString converts a slice of Nodes to a string, separated by sep.
 func sliceToString[T Node](s []T, sep string) string {
-	nodeStrings := make([]string, len(s))
-	for i, node := range s {
-		nodeStrings[i] = fmt.Sprint(node)
+	if len(s) == 0 {
+		return ""
 	}
 
-	return strings.Join(nodeStrings, sep)
+	r := s[0].String()
+	for _, v := range s[1:] {
+		r += sep + v.String()
+	}
+
+	return r
 }
--- a/ast.go
+++ b/ast.go
@@ -253,7 +253,7 @@ type PrivateUseAnnotation struct {
 	// TODO: Implementation
 }
 
-func (PrivateUseAnnotation) String() string { return "^ PRIVATE USE ANNOTATION NOT IMPLEMENTED" } // TODO: Implement
+func (PrivateUseAnnotation) String() string { return "^ PRIVATE_USE_ANNOTATION_NOT_IMPLEMENTED" } // TODO: Implement
 
 type ReservedAnnotation struct {
 	Annotation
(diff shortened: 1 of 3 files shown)


ast.go
```diff
@@ -1,7 +1,33 @@
...
+		Message: SimpleMessage{
+			Patterns: []Pattern{
+				TextPattern("Hello, "),
+				PlaceholderPattern{Expression: VariableExpression{Variable: Variable("variable")}},
+				TextPattern(" World!"),
+			},
+		},
+	}
+
+	fmt.Print(ast) // Hello, { $variable } World!
```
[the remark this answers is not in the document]
VladislavsPerkanuks:
```abnf
variable-expression = "{" [s] variable [s annotation] [s] "}"
; where s is whitespace and [] is optional, so [s] == optional whitespace
```
so both options are correct
```
Hello, { $variable } World!
Hello, {$variable} World!
```

ast.go
```diff
@@ -111,43 +161,74 @@ type LiteralExpression struct {
...
 
 type NumberLiteral float64
 
-func (NumberLiteral) node()     {}
-func (NumberLiteral) literal()  {}
-func (NumberLiteral) unquoted() {}
+func (NumberLiteral) node()             {}
+func (NumberLiteral) literal()          {}
+func (NumberLiteral) unquoted()         {}
+func (nl NumberLiteral) String() string { return fmt.Sprint(float64(nl)) }
```
klavinsjanis, line 231: is type conversion to float64 necessary?

VladislavsPerkanuks:
```go
fmt.Sprint(nl)
```

This one will go into infinite recursion :)

parse_test.go
```diff
@@ -327,7 +328,22 @@ func TestParseSimpleMessage(t *testing.T) {
 			actual, err := Parse(tt.input)
 			require.NoError(t, err)
 
-			require.Equal(t, tt.expected, actual)
+			// Check that AST message is equal to expected one.
+			require.Equal(t, tt.expected, actual.Message)
+
+			// Check that AST message converted back to string is equal to input.
+
+			// Edge case: scientific notation number is converted to normal notation, hence comparison is bound to fail.
```
klavinsjanis, line 336: I wonder if there is any way to preserve the notation

VladislavsPerkanuks: in token we store string representation of numberLIteral
in node we store float representation of numberLiteral, using json.Unmarshal

so json.Unmarshal converts "1e3" to 1000.0

going back to string we don't have any information about the original notation.

So simple answer: it is not possible

klavinsjanis: Maybe we could store it as a string in node?

ast.go
```diff
@@ -248,16 +369,52 @@ type Identifier struct {
...
+
+// sliceToString converts a slice of Nodes to a string, separated by sep.
+func sliceToString[T Node](s []T, sep string) string {
+	nodeStrings := make([]string, len(s))
+	for i, node := range s {
+		nodeStrings[i] = fmt.Sprint(node)
+	}
+
+	return strings.Join(nodeStrings, sep)
+}
```
janishorsts, line 420: more efficient

```suggestion
func sliceToString[T Node](s []T, sep string) string {
	if len(s) == 0 {
		return ""
	}
	
	r := fmt.Sprint(s[0])
	
	for _, v := range s[1:] {
		r += sep + fmt.Sprint(v)
	}
	
	return r
}
```

ast.go
```diff
@@ -158,18 +239,30 @@ type FunctionAnnotation struct {
...
+	return fmt.Sprintf("%s %s", fa.Function, sliceToString(fa.Options, " "))
+}
+
 type PrivateUseAnnotation struct {
 	Annotation
 
 	// TODO: Implementation
 }
 
+func (PrivateUseAnnotation) String() string { return "TODO" }
```
janishorsts, line 256: This breaks MF2 syntax. Use something like `.|private annotation not implemented|`.

janishorsts (approved):

DarjaLeonova (approved):

klavinsjanis (approved):

PR #14, opened 2023-12-18
AST Validate initial implementation

VladislavsPerkanuks:
Check if AST tree complies with ABNF syntax

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -1,6 +1,7 @@
 package mf2
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 )
@@ -11,7 +12,22 @@ type AST struct {
 }
 
 // String returns the string representation of the AST, i.e. MF2 formatted message.
-func (ast AST) String() string { return fmt.Sprint(ast.Message) }
+func (a AST) String() string { return fmt.Sprint(a.Message) }
+
+// Validate returns an error if the AST is invalid according to the MessageFormat 2.0 specification.
+// For example, when matcher has no selectors or variants.
+// Or variable is zero value, i.e $.
+func (a AST) Validate() error {
+	if a.Message == nil {
+		return errors.New("ast: message is required")
+	}
+
+	if err := a.Message.Validate(); err != nil {
+		return fmt.Errorf("ast.%w", err)
+	}
+
+	return nil
+}
 
 // --------------------------------Interfaces----------------------------------
 //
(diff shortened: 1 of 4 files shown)


janishorsts:
Parse() should produce a valid AST. 

What is the reason or use case to have explicit Validate() for validation?

VladislavsPerkanuks:
The primary role of the Parser is to transform lexer tokens into an `AST`.
It should verify the correct sequence of tokens, for instance, a variable token or a literal token should follow an expression token, not random tokens. The validity of a Node **may or may not be** checked.

This leads me to propose two approaches:

### Approach 1:
During the `Parse()` function, when a node is created, its validity is also checked. For instance, while parsing a variable, it verifies if the variable name is not empty, like `$`. This applies to every potentially problematic node, such as matchers with no variants or selectors, empty literals, etc.

### Approach 2:
We validate after parsing all nodes. An explicit` Validate()` function is used in `Parse() `to check if the AST is valid.

I prefer the second approach for the following reasons:

- It separates the validation implementation from the parsing implementation, simplifying the parsing process.
- In the future, when we construct AST trees, either manually or with a builder, it could validate it. This wouldn’t be possible with the first approach.

janishorsts:
@VladislavsPerkanuks, validation must be part of `Parse()` and not exposed. As a consumer, I expect `Parse()` to produce a valid AST or error if it is not a valid MF2 syntax.

ast.go
```diff
@@ -184,6 +280,22 @@ func (ve VariableExpression) String() string {
 	return fmt.Sprintf("{ %s %s }", ve.Variable, ve.Annotation)
 }
 
+func (ve VariableExpression) validate() error {
+	if err := ve.Variable.validate(); err != nil {
```
mvilks, line 284: no need to check for ve.Variable being nil?

mvilks (approved):

janishorsts (approved):
I would prefer validating while reading the tokens.

This ensures the syntax is correct.

Post-validation does not ensure it, only that the final state is correct.

DarjaLeonova (approved):

ast.go
VladislavsPerkanuks: underlying value is string, cant be nil

VladislavsPerkanuks:
Syntax validation is already done:
some validation is in lexer
some validation is in parser.

after parser error handling PR merge, it checks, if next token is valid in given context. E.g. if in expression there is no text tokens etc. 
so syntax is checked.

This PR checks, that final state is correct, when we know that syntax(tokens and their order) is correct.

So in summary: after this PR
If there were no error during Parse(), the AST tree is 100% correct and ready to be used

PR #19, opened 2023-12-27
Lexer: Break down itemKeyword into multiple items

### LINKED ISSUE #16: Lexer: Break down keyword item into multiple items
For now, we have one item for all keywords, `match, input, ...`
That makes parsing matcher and declarations harder compared to other elements

```go
// now
switch itm.typ {
  case itemKeyword:
    switch itm.val {
      case "." + keywordInput: // Input declaration
      case "." + keywordLocal: // Local declaration
      ...
  }
}

// want
switch itm.typ {
  case itemInputKeyword:
    // parse InputDeclaration
  case itemlocalKeyword:
    // parse LocalDeclaration
...
}
```

```diff
const (
...
- itemKeyword
+ itemInputKeyword
+ itemLocalKeyword
+ itemMatchKeyword
+ itemReservedKeyword
...
)
```

VladislavsPerkanuks:
closes #16

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,52 +1,52 @@
 linters:
   enable-all: true
   disable:
-   - asasalint
-   - deadcode
-   - depguard
-   - dupl
-   - exhaustivestruct
-   - exhaustruct
-   - forbidigo
-   - funlen
-   - gci
-   - ginkgolinter
-   - goconst
-   - gochecknoglobals
-   - gochecknoinits
-   - gocyclo
-   - godox
-   - goerr113
-   - gofmt
-   - goheader
-   - goimports
-   - golint
-   - gomodguard
-   - gosmopolitan
-   - grouper
-   - ifshort
-   - inamedparam
-   - interfacer
-   - loggercheck
-   - maintidx
-   - maligned
-   - nlreturn
-   - nonamedreturns
-   - nosnakecase
-   - promlinter
-   - revive
-   - scopelint
-   - structcheck
-   - tagliatelle
-   - testableexamples
-   - testpackage
-   - thelper
-   - tparallel
-   - varcheck
-   - varnamelen
-   - wastedassign
-   - whitespace
-   - zerologlint
+    - asasalint
+    - deadcode
+    - depguard
+    - dupl
+    - exhaustivestruct
+    - exhaustruct
+    - forbidigo
+    - funlen
+    - gci
+    - ginkgolinter
+    - goconst
+    - gochecknoglobals
+    - gochecknoinits
+    - gocyclo
+    - godox
+    - goerr113
+    - gofmt
+    - goheader
+    - goimports
+    - golint
+    - gomodguard
+    - gosmopolitan
+    - grouper
+    - ifshort
+    - inamedparam
+    - interfacer
+    - loggercheck
+    - maintidx
+    - maligned
+    - nlreturn
+    - nonamedreturns
+    - nosnakecase
+    - promlinter
+    - revive
+    - scopelint
+    - structcheck
+    - tagliatelle
+    - testableexamples
+    - testpackage
+    - thelper
+    - tparallel
+    - varcheck
+    - varnamelen
+    - wastedassign
+    - whitespace
+    - zerologlint
 linters-settings:
   cyclop:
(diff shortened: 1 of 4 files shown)


parse.go
```diff
@@ -45,6 +45,17 @@ func (p *parser) collect() error {
...
 
+// isComplexMessage returns true if first token is one of the complex message tokens.
+func (p *parser) isComplexMessage() bool {
+	//nolint:exhaustive
+	switch p.items[0].typ {
+	case itemInputKeyword, itemLocalKeyword, itemMatchKeyword, itemReservedKeyword, itemQuotedPatternOpen:
+		return true
+	}
+
+	return false
```
janishorsts, line 56:
```suggestion
	switch p.items[0].typ {
	default:
		return false
	case itemInputKeyword, itemLocalKeyword, itemMatchKeyword, itemReservedKeyword, itemQuotedPatternOpen:
		return true
	}
```

parse.go
```diff
@@ -151,39 +162,25 @@ func (p *parser) parseComplexMessage() (ComplexMessage, error) {
...
-				declarations = append(declarations, declaration)
 
-			case "." + keywordLocal: // Local declaration
-				p.next() // skip keyword
-
-				declaration, err := p.parseLocalDeclaration()
-				if err != nil {
-					return ComplexMessage{}, fmt.Errorf("parse local declaration: %w", err)
-				}
+		case itemInputKeyword, itemLocalKeyword, itemReservedKeyword: // Declarations
```
janishorsts, line 166: Why did you add `parseDeclaration()`? Instead. as you suggested per the ticket?

VladislavsPerkanuks: because error handling makes it ugly
originlly when ticket was created error handling was not yet implemented, so
that would be simply
```go
switch itm.typ {
  case itemInputKeyword:
    declarations = append(declarations, p.parseInputDeclaration())
  case itemlocalKeyword:
    declarations = append(declarations, p.parseLocalDeclaration())
...
}
```

but now it would be

janishorsts: what do you mean by ugly?

janishorsts: by ugly, do you mean the `if` statement?

```go
case itemInputKeyword:
  v, err := p.parseInputKeyword()
  if err != nil {
  /// ...
  }
  
  declarations = append(declarations, v)

Comment (VladislavsPerkanuks) [reply parse.go:166]: yeah, for every declaration type

Comment (VladislavsPerkanuks) [reply parse.go:166]: i can change it no problem, but I just thought, that is a little bit cleaner that way

Comment (janishorsts) [reply parse.go:166]: I'm not convinced this PR improves the code:

* the code size increased significantly
* cognitive load increased as well

Is it worth still to pursue it? Should we postpone improvement until we figure out how to do it properly?

Comment (VladislavsPerkanuks) [reply parse.go:166]: code size increasement is mainly due to exhaustive linter. I would drop it as my experience it gives more headache than benefits :)

Otherwise apart from that, there is no significant code size changes.

About cognitive load, it decreased in one way, but maybe increased in other way, so I would say that is not changed much

File: parse.go [hunk 3]
```diff
@@ -234,8 +254,10 @@ func (p *parser) parsePatterns() ([]Pattern, error) {
 			pattern = append(pattern, PlaceholderPattern{Expression: expression})
 		// bad tokens
 		case itemEOF, itemVariable, itemFunction,
-			itemExpressionClose, itemQuotedPatternOpen, itemQuotedPatternClose,
-			itemKeyword, itemCatchAllKey,
+			itemExpressionClose, itemQuotedPatternOpen,
```
janishorsts, line 257: Why not `default`?

janishorsts: yeah, all other tokens are bad - `default` :)

.golangci.yml
```diff
@@ -1,52 +1,52 @@
...
+    - tagliatelle
+    - testableexamples
+    - testpackage
+    - thelper
+    - tparallel
+    - varcheck
+    - varnamelen
+    - wastedassign
+    - whitespace
+    - zerologlint
```
VladislavsPerkanuks, line 49: 3 spaces -> 4 spaces

janishorsts (approved):

PR #20, opened 2023-12-28
12-builder

klavinsjanis:


builder/builder.go
```diff
@@ -0,0 +1,390 @@
...
+
+	return b
+}
+
+func (b *Builder) Text(s string) *Builder {
+	if b.err != nil {
+		return b
+	}
+
+	s = escapeSpecialChars(s)
```
VladislavsPerkanuks, line 98: not correct

for text only three characteres should be escaped
```abnf
text-escape     = backslash ( backslash / "{" / "}" )
```

builder/builder.go
```diff
@@ -0,0 +1,390 @@
...
+	'@': {},
+	'#': {},
+	'%': {},
+	'*': {},
+	'<': {},
+	'>': {},
+	'/': {},
+	'?': {},
+	'~': {},
+}
```
VladislavsPerkanuks, line 360: for every node type there a different escapes that needs to be done
```abnf
text-escape     = backslash ( backslash / "{" / "}" ) ; for text node
quoted-escape   = backslash ( backslash / "|" ) ' ; for quoted literal
reserved-escape = backslash ( backslash / "{" / "|" / "}" ) ; for reserved statement
```

builder/builder.go
```diff
@@ -0,0 +1,390 @@
...
+
+type function struct {
+	name    string
+	options []option
+}
+
+type Expression struct {
+	spacing   string
+	operand   any // literal or variable
+	functions []function
```
VladislavsPerkanuks, line 253: Expression can have only one function
```abnf
annotation = (function *(s option))
```

\+ and \- functions, are two separate expressions
hello {+func1} world {-func2}

builder/builder_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Parallel()
+
+	for _, test := range []struct { //nolint:govet
+		name, expected string
+		b              *Builder
+	}{
+		{
+			"text",
+			"Hello, World\\!",
+			New().Text("Hello, World!"),
```
VladislavsPerkanuks, line 19: bad practice, populating structs without field names

builder/builder.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func New() *Builder {
+	return &Builder{
+		newline: "\n",
+		spacing: " ",
+	}
+}
+
+func (b *Builder) Build() (string, error) {
```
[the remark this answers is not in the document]
janishorsts: Not exactly. `strings.Builder` provide safe string concatenation.

Unless we allow consumers to build broken MF2.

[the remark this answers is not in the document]
janishorsts: The target consumers and intent are different. We don't want consumers to use AST to build MF2 messages.

* Parse() parses MF2 into AST (we don't have the intent to do AST.String(), we did it by inertia)
* Builder makes it easy and safe to construct MF2 messages

E.g. we don't want to burden consumers on how to handle quoted/unquoted literals. The consumer does NOT care.

Builder helps us to convert from different i18n formats into MF2.

builder/builder.go
```diff
@@ -0,0 +1,390 @@
...
+	default:
+		panic(fmt.Sprintf("unsupported literal type: %T", v))
+	}
+}
+
+// helpers
+
+// specialChars is a list of characters that need to be escaped in the node.Text,
+// because they are either syntax characters or reserved keywords.
+var specialChars = map[rune]struct{}{
```
janishorsts, line 343: Can we please stop using maps for static list checking in Golang? Please use `switch` instead. 

Also, please read @VladislavsPerkanuks comment on special characters. This is old stuff.

builder/builder.go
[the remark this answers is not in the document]
VladislavsPerkanuks: its not about that.

like AST, has Stringer implementation, right ? AST -> mf2 string.

my question is:
why cant builder have additional field unexported ast, to call that AST Stringer implementation, instead of defining new Stringer implementation for those new structs, which are doing the same thing

builder/builder.go
janishorsts: How do you handle the case when a raw text starts (MF2 message) with `.`, e.g. `.ok.`? It should be a complex message `{{.ok.}}`.

Is it more appropriate to handle serialization in the `Build()` method?

VladislavsPerkanuks: if mf2 string is  ".ok."
then it is simple message

janishorsts: I'm afraid it is NOT.

```abnf
simple-message = [simple-start pattern]
simple-start = simple-start-char / text-escape / placeholder
simple-start-char = %x0-2D         ; omit . <--------
````

Comment (janishorsts) [reply builder/builder.go:98]: simple message MUST start with `simple-start-char`

Comment (VladislavsPerkanuks) [reply builder/builder.go:98]: i see,
then it cannot start with `.`

Comment (VladislavsPerkanuks) [reply builder/builder.go:98]: it simply cannot start with `.`
neither in simpleMessage neither in ComplexMessage

so `{{.ok.}}` is also invalid mf2 syntax

Comment (VladislavsPerkanuks) [reply builder/builder.go:98]: okay nevermind it can :)

File: builder/builder.go [hunk 7]
```diff
@@ -0,0 +1,386 @@
...
+	default:
+		panic(fmt.Sprintf("unsupported literal type: %T", v))
+	}
+}
+
+// helpers
+
+// quotedEscape escapes special characters in quoted name literal.
+func quotedEscape(s string) string {
+	var sb strings.Builder
```
VladislavsPerkanuks, line 354: strings.Replacer is cleaner and more straigforward

builder/builder.go
```diff
@@ -0,0 +1,386 @@
...
+
+		sb.WriteRune(c)
+	}
+
+	return sb.String()
+}
+
+// textEscape escapes special characters in text.
+func textEscape(s string, simpleStart bool) string {
+	var sb strings.Builder
```
VladislavsPerkanuks, line 370: strings.Replacer is cleaner and more straigforward

builder/builder.go
```diff
@@ -0,0 +1,386 @@
...
+
+func printLiteral(l any) string {
+	switch v := l.(type) {
+	case string:
+		v = quotedEscape(v)
+		return "|" + v + "|"
+	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
+		return fmt.Sprintf("|%d|", v)
+	case float32, float64:
+		return fmt.Sprintf("|%f|", v)
```
VladislavsPerkanuks, line 344: numeric literals are unquoted

builder/builder.go
```diff
@@ -0,0 +1,386 @@
...
+		e.function = function{name: name, options: option}
+		return e
+	}
+}
+
+func printLiteral(l any) string {
+	switch v := l.(type) {
+	case string:
+		v = quotedEscape(v)
+		return "|" + v + "|"
```
VladislavsPerkanuks, line 340: not all literals ar quoted

builder/builder_test.go
```diff
@@ -0,0 +1,95 @@
...
+			"Hello { +link } World { -link }",
+			New().
+				Text("Hello ").
+				Expr(Expr().Func("+link")).
+				Text(" World ").
+				Expr(Expr().Func("-link")),
+		},
+		{
+			"local",
+			".local $hostName = { $host }\n",
```
VladislavsPerkanuks, line 58: Isn't the idea of a builder to construct, a valid mf2 syntax message, and error if something is missing/incorrect?

Because this is not a valid complex message

builder/builder.go
klavinsjanis: I know, but we are using single method Literal() for now.

builder/builder.go
klavinsjanis: is it going to be faster?

builder/builder_test.go
klavinsjanis: ok, will look into this one. Currently, we are only focusing on the happy path.

builder/builder.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+type local struct {
+	expr     *Expression
+	variable variable
+}
+
+type literal any
+
+type variable string
```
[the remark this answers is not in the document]
klavinsjanis: I am not really sure about this one, maybe it's there for consistency.

builder/builder.go
janishorsts: Complex message can start

```abnf
text-char = simple-start-char / "."
```

builder/builder.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+func (b *Builder) String() string {
+	s, err := b.Build()
+	if err != nil {
+		panic(err)
+	}
+
+	return s
+}
```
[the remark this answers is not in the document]
janishorsts: @klavinsjanis please remove `String()`, I added it temporarily.

builder/builder.go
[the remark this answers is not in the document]
VladislavsPerkanuks: my main concern is the following:
we have almost perfect nodes defined in AST, with each node, representing itself as string. But here instead of using those nodes, we are creating derivatives of them, almost the same but a little different. 
My main concerns:
a) Instead of maintaining only ast.go nodes, we must also maintain builder.go nodes, in case the messageformat abnf changes
b) Instead of using already implemented node->string, we are reimplementing the same thing for derivativeNode -> string, which strengthens my point a)

builder/builder_test.go
janishorsts: @VladislavsPerkanuks please state what is not correct! :)

According to my knowledge and ABNF spec, it is correct.

builder/builder.go
VladislavsPerkanuks: `readability > speed` almost in any scenario

builder/builder_test.go
VladislavsPerkanuks:
```abnf
```

builder/builder.go
```diff
@@ -0,0 +1,386 @@
...
+type literal any
+
+type variable string
+
+type function struct {
+	name    string
+	options []option
+}
+
+type Expression struct {
```
VladislavsPerkanuks, line 255: why exported ?

builder/builder.go
janishorsts: @klavinsjanis, based on the type and content, you have to figure out how to print it - quoted or unquoted.

builder/builder.go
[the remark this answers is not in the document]
VladislavsPerkanuks: in AST.go there is a type variable with an underlying type of string, but it is done to implement interfaces.
Here it is not needed, and it just additional headache because you need to type cast

builder/builder.go
```diff
@@ -0,0 +1,569 @@
...
+}
+
+func New() *Builder {
+	return &Builder{
+		newline: "\n",
+		spacing: " ",
+	}
+}
+
+func (b *Builder) build() (string, error) {
```
janishorsts, line 27:
```suggestion
func (b *Builder) Build() (string, error) {
```

builder/builder.go
```diff
@@ -0,0 +1,569 @@
...
+func (b *Builder) Text(s string) *Builder {
+	if b.err != nil {
+		return b
+	}
+
+	if b.IsEmpty() {
+		if hasSimpleStart(s) || s == "" {
+			b.patterns = append(b.patterns, textEscape(s))
+		} else {
+			b.quoted = []any{textEscape(s)} // create quoted pattern if the first character doesn't have simple start.
```
janishorsts, line 148: Why do we care about escaped text at this stage? It is the responsibility of the `Build()` to figure out how to serialize it, right?

I don't think we need `quoted` field. Why not simply append it to `patterns`?

builder/builder.go
```diff
@@ -0,0 +1,569 @@
...
+		0x2F <= r && r <= 0x5B || // omit \
+		0x5D <= r && r <= 0x7A || // omit {}
+		r == 0x7C || // omit }
+		0x7E <= r && r <= 0xD7FF || // omit surrogates
+		0xE000 <= r && r <= 0x10FFFF
+}
+
+// IsEmpty returns true if the builder is empty.
+func (b Builder) IsEmpty() bool {
+	if b.locals == nil && b.inputs == nil && b.selectors == nil &&
```
janishorsts, line 544: **Never** check if the slice is empty by comparing it to nil.

IT IS BAD, and causes issues.

Nil vs empty slices.

builder/builder.go
```diff
@@ -0,0 +1,569 @@
...
+}
+
+func (b *Builder) Text(s string) *Builder {
+	if b.err != nil {
+		return b
+	}
+
+	if b.IsEmpty() {
+		if hasSimpleStart(s) || s == "" {
+			b.patterns = append(b.patterns, textEscape(s))
```
janishorsts, line 146: Please serialize values in the `Build()`. Do not do it early. Makes it harder to read and maintain the code.

`Build()` - concatenates and serializes values, and builds final value.

builder/builder.go
klavinsjanis: I have a custom type patterns for quoted, that includes custom methods.

builder/builder.go
janishorsts: I mean, use raw values in build steps.

builder/builder.go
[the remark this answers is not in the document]
janishorsts: Not sure, why we have `String()` in AST.  Do we have it? AST is for parsing and building AST, and its responsibility is **NOT** to serialize AST into MF2 syntax.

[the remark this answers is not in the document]
VladislavsPerkanuks: AST -> String is a convenient way to convert AST nodes to string, for saving in DB.

**Scenario:**
We want to update message that is stored in our DB
1. We get MF2 string from DB
2. We call `Parse() `to get AST from String
3. We make changes to AST
4. We save AST.String(), into our DB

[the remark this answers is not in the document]
VladislavsPerkanuks: As in our current workflow in translation project, translation works as following:
we extract only translatable text from MF2 AST, translate it, and then replace back into AST.

**So in that Scenario:**
1. We get MF2 string from DB
2. We call `Parse()` to get AST
3. We get only translatable text from AST. (i.e only text nodes)
4. We translate it
5. We replace translatable text nodes with translated text nodes
6. We save `AST.String()`, into DB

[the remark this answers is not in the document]
janishorsts: I'd like to separate all the different concerns. The Builder is for different use cases.

We have 3 responsibilities in the `mf2` repo:
* parsing
* building
* executor (not implemented yet)

If all is good, we'll use AST nodes in the builder's implementation without exposing the implementation.

builder/builder.go
VladislavsPerkanuks: image

strings.Replacer is faster

janishorsts: @VladislavsPerkanuks can you add benchtest for inline string concat?

builder/builder.go
VladislavsPerkanuks: but why tho ?

builder/builder.go
```diff
@@ -0,0 +1,550 @@
...
+		panic(fmt.Sprintf("function MUST start with :, + or -: %s", name))
+	case ':', '+', '-':
+		e.function = function{name: name, options: option}
+		return e
+	}
+}
+
+func printLiteral(l any) string {
+	switch v := l.(type) { // TODO: more liberal
+	case string:
```
VladislavsPerkanuks, line 394: so 2/3 literal cases are covered
this can print number, and quoted literals, what about unquoted ? it gets ignored ?

builder/builder.go
klavinsjanis: Exposed functions must return exposed types apparently.
func Expr() *Expression { return new(Expression) }

builder/builder.go
klavinsjanis: prints unqouted if `name = name-start *name-char` holds true

builder/builder.go
VladislavsPerkanuks: okay

builder/builder.go
VladislavsPerkanuks: but i see here only two returns ?
when 3 expected

klavinsjanis: string -> unqouted/quoted
number -> unqouted
unknown -> panic

VladislavsPerkanuks: i see now

PR #22, opened 2024-01-02
Lexer: Omit Keywords and Variables Prefixes

### LINKED ISSUE #9: Lexer: Variable token should omit `$`
```go
item{typ: itemVariable, val: "$variable"} // current
item{typ: itemVariable, val: "variable"}  // expected
```

VladislavsPerkanuks:
closes #9 

~**waiting for #19 as conflicts are expected**~

**Also**:
- Omit `.` prefixes for keywords
- Refactor `lexName` to reflect changes + readability improvements?


lex.go
```diff
@@ -370,32 +370,33 @@ func lexExpr(l *lexer) stateFn {
...
+		l.backup() // backup to the first rune
+	}
 
-		switch {
-		default:
-			l.backup()
+	var (
+		s string // item value
+		r rune   // current rune
+	)
```
janishorsts, line 389:
```suggestion
	var s string // item value
```

lex.go
```diff
@@ -370,32 +370,33 @@ func lexExpr(l *lexer) stateFn {
...
-		case len(s) == 0 && isNameStart(r):
-			s = string(r)
-		case len(s) == 0 && r == '$':
-			s = string(r)
-			typ = itemVariable
-		case len(s) == 0 && r == '.':
-			s = string(r)
-			typ = itemKeyword
-		}
+	for r = l.next(); isName(r); r = l.next() {
```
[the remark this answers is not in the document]
VladislavsPerkanuks: cannot, r is needed outside the loop

lex.go
VladislavsPerkanuks: cannot do that, r is needed outside the loop

lex.go
```diff
@@ -370,32 +370,33 @@ func lexExpr(l *lexer) stateFn {
...
+		s += string(r)
 	}
+
+	if r != eof { // backup is not necessary if we reached eof
```
janishorsts, line 395: Isn't `eof` broken MF2 syntax?

Name is either a variable name or an unquoted literal.

VladislavsPerkanuks: it is, but it has same behaviour as it was before

lex.go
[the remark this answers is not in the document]
janishorsts: what about this?

```suggestion
	for r := l.next(); isName(r); r = l.next() {
	  if r == eof {
	    // unexpected eof
	  }
	}
	
	l.backup()
	
	// ...
```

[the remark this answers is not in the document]
janishorsts: or what you have but error on eof

```go
if r == eof {
 // error on unexpected eof
}

l.backup()
```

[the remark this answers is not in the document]
VladislavsPerkanuks: this could work, previous one I think will not, as eof will fail isName().

lex.go
janishorsts: Let's do it. Let's return the unexpected eof error.

janishorsts (approved):

PR #27, opened 2024-01-10
24-builder-prefixes

DarjaLeonova:


builder.go
```diff
@@ -317,7 +318,7 @@ func (e *Expression) build(spacing string) string {
 		for _, o := range e.function.options {
 			s += " " + o.key + spacing + "=" + spacing
 
-			if v, ok := o.operand.(string); ok && len(v) > 0 && v[0] == '$' { // recognized as variable if string starts with $
+			if v, ok := o.operand.(string); ok && len(v) > 0 {
```
klavinsjanis, line 321: It might fail on quoted literals after removal of - `&& v[0] == '$'`

janishorsts: Also, the option value is not optional. E.g., an empty string should be `host = ||`.

builder.go
```diff
@@ -175,7 +176,7 @@ func (b *Builder) Local(v string, expr *Expression) *Builder {
 		return b
 	}
 
-	b.locals = append(b.locals, local{variable: variable(v), expr: expr})
+	b.locals = append(b.locals, local{variable: variable(varSymbol + v), expr: expr})
```
janishorsts, line 179: I would prefer to prefix '$' only when serializing/building to MF2 syntax.

builder.go
```diff
@@ -361,27 +358,45 @@ type FuncOption struct {
...
 }
 
-func Option(key string, operand any) FuncOption { return FuncOption{key: key, operand: operand} }
-
 func Func(name string, option ...FuncOption) *Expression {
 	return Expr().Func(name, option...)
 }
 
+func VarOption(name, varName string) FuncOption {
+	return FuncOption{key: name, operand: varSymbol + varName}
```
janishorsts, line 366: We use `variable` type in the `func (e *Expression) Var(name string)` but not here. Why?

It should be consistent - we cannot use `string` in one place and `variable` in another for the same type.

```suggestion
	return FuncOption{key: name, operand: variable(varName)}
```

builder.go
```diff
@@ -338,20 +339,16 @@ func (e *Expression) Literal(v any) *Expression {
...
+	if len(name) == 0 {
 		panic("variable name cannot be empty")
 	}
 
-	if v[0] != '$' {
-		panic(fmt.Sprintf("variable must start with $: %s", v))
-	}
-
-	e.operand = variable(v)
+	e.operand = variable(varSymbol + name)
```
VladislavsPerkanuks, line 351: same as @janishorsts mentioned. `$` symbol is used only for serialization, so we can skip it here

klavinsjanis:
Would be good to see tests with the examples mentioned in the previous comments.

builder.go
DarjaLeonova: We don't need to write the dollar symbol before the variable anymore, so why will it fail?

klavinsjanis (approved):

builder.go
VladislavsPerkanuks, line 179: i thought we quit on adding prefixes on variables, and literals
```suggestion
	b.locals = append(b.locals, local{variable: variable(v), expr: expr})
```

builder.go
```diff
@@ -338,20 +339,16 @@ func (e *Expression) Literal(v any) *Expression {
...
 
-func Var(s string) *Expression {
-	return Expr().Var(s)
+func Var(name string) *Expression {
+	return Expr().Var(name)
 }
 
-func (e *Expression) Var(v string) *Expression {
-	if len(v) == 0 {
+func (e *Expression) Var(name string) *Expression {
```
VladislavsPerkanuks, line 346: please add checking if variable is valid. according to [abnf](https://github.com/unicode-org/message-format-wg/blob/7c00820a0462679eba696181c45bfadb43d2eedd/spec/message.abnf#L88-L95)

e.g. 
calling Var with argument "$var" should error/panic

builder.go
```diff
@@ -361,33 +358,51 @@ type FuncOption struct {
 	key     string
 }
 
-func Option(key string, operand any) FuncOption { return FuncOption{key: key, operand: operand} }
+func VarOption(name, varName string) FuncOption {
+	return FuncOption{key: name, operand: variable(varName)}
```
VladislavsPerkanuks, line 362: same 
add checking if variable is valid. according to [abnf](https://github.com/unicode-org/message-format-wg/blob/7c00820a0462679eba696181c45bfadb43d2eedd/spec/message.abnf#L88-L95)

builder.go
DarjaLeonova: Yes, you are right, forgot to remove it ;)

builder.go
```diff
@@ -46,7 +48,7 @@ func (b *Builder) Build() (string, error) {
 	}
 
 	for _, v := range b.locals {
-		s += ".local" + b.spacing + string(v.variable) + b.spacing + "=" + b.spacing + v.expr.build(b.spacing) + b.newline
+		s += ".local" + b.spacing + varSymbol + string(v.variable) + b.spacing + "=" + b.spacing + v.expr.build(b.spacing) + b.newline //nolint:lll
```
janishorsts, line 51: It's ok to break this line

```suggestion
		s += ".local" + b.spacing + varSymbol + string(v.variable) + b.spacing + "=" +
		b.spacing + v.expr.build(b.spacing) + b.newline
```

builder.go
```diff
@@ -544,3 +564,13 @@ func coalesce[T comparable](l ...T) T {
 
 	return c
 }
+
+// validateVarName checks whether the first rune is a valid starting character
+// for a variable name according to the specified criteria.
+func validateVarName(varName string) {
```
janishorsts, line 570: Remove the validation - I'd rather have no validation than an incomplete one, it validates only the first character.

Two issues:
* panics, interrupts the build flow - we store error in the Builder struct
* incomplete validation.

VladislavsPerkanuks (approved):

PR #28, opened 2024-01-11
24 - improve builder prefixes

janishorsts:


builder_test.go
```diff
@@ -72,70 +72,70 @@ func Test_Builder(t *testing.T) {
...
-					Var("$i"),
-					Var("$j"),
+					Expr().Var("i"),
+					Expr().Var("j"),
 				).
 				Keys(1, 2).Text("{first}").
-				Keys(2, 0).Text("second ").Expr(Var("$i")).
-				Keys(3, 0).Expr(Literal("\\a|")).
-				Keys("*", "*").Expr(Literal(1)).
+				Keys(2, 0).Text("second ").Expr(Expr().Var("i")).
```
klavinsjanis, line 136: Hmm `Expr(Expr().Var("i")).` this doesn't look too good.

janishorsts:
```go
// it is already possible, both ways are correct
Expr(Var("i")
Expr().Var("i")

// literal - no need to specify type, literal implicitly
Expr("Hello")
Expr(Literal("Hello"))
Expr().Literal("Hello")
```

janishorsts:
Revisit it at a later time. Not relevant at this stage.

PR #32, opened 2024-01-22
26-escapes-in-quoted-literal-breaks-lexer

DarjaLeonova:


kaaposc (approved):

VladislavsPerkanuks (approved):

klavinsjanis (approved):

parse/lex_test.go
```diff
@@ -320,6 +320,15 @@ func Test_lex(t *testing.T) {
...
 		},
+		{
+			name:  "quoted literal with escapes",
+			input: "{|Here this is escaped \\| it is part of literal|}", // Here is escaped | it is part of literal
+			expected: []item{
+				mk(itemExpressionOpen, "{"),
+				mk(itemQuotedLiteral, "Here this is escaped | it is part of literal"),
+				mk(itemExpressionClose, "}"),
+			},
+		},
```
janishorsts, line 331: Update the `quoted literal` test with the input like `"{|\\| is escaped| :uppercase}"`.

When lexing MF2, creating a test for each permutation is quite expensive - the number of tests we can create is limitless. Less is better. :)

```suggestion
```

janishorsts (approved):

PR #33, opened 2024-01-23
Lexer: Private Annotation

VladislavsPerkanuks:


.vscode/settings.json
```diff
@@ -2,5 +2,11 @@
   "go.lintOnSave": "workspace",
   "go.coverOnSave": true,
   "go.testOnSave": true,
-  "go.lintTool": "golangci-lint"
+  "go.lintTool": "golangci-lint",
+  // When running with latest dlv
```
janishorsts, line 6: Does it throw an error if this is not added? What does it do?

VladislavsPerkanuks: throws an error when debugging
says: minimal supported go version is 1.19

VladislavsPerkanuks: Usually i downgraded dlv if I wanted to debug, but this seems to work fine too

klavinsjanis (approved):

DarjaLeonova (approved):

.vscode/settings.json
janishorsts: Why downgrade dlv to debug? Can you provide more details?

What are the side effects of disabling it? Can disabling it send us into a rabbit hole when we face an issue we cannot explain?

VladislavsPerkanuks: when i work with this project i use go 1.18. Using latest dlv, it throws that this version of go is not supported.
so options are
1) add this flag to dlv
2) downgrade dlv 

it seems that debugger works fine with this flag, and it is easier than going back and forth with dlv releases 🤷

janishorsts: Works for me - I installed the dlv@latest and used go1.21.6, and debugging works as expected.

VladislavsPerkanuks: it is really no big dealie

janishorsts: I see.

My advice is `always use the latest Golang`.

There are no major benefits of using the min Golang required by a project. Let the CI deal with the support of older Golang.

I have always used the latest Golang in all projects, even with super old Golang versions. `It all works just fine`.

There are a few challenges, but they are edge cases and super rare.

You may discover that you used something from the standard packages that is not available in the older versions. Refactoring is always easy, and not an obstacle.

.vscode/settings.json
```diff
@@ -2,5 +2,11 @@
...
+  "go.delveConfig": {
+    "dlvFlags": [
+      "--check-go-version=false"
+    ]
+  },
```
janishorsts, line 11: This can be removed.

```suggestion
```

Simply use the latest Golang!

.vscode/settings.json
VladislavsPerkanuks: sure

janishorsts (approved):

PR #38, opened 2024-01-30
Reduce AST complexity: Message

VladislavsPerkanuks:


parse/parse_test.go
VladislavsPerkanuks, line 1: Check, the hide whitespaces mark, then it Diff is much cleaner
image

DarjaLeonova (approved):

PR #41, opened 2024-01-30
Reduce AST complexity: Literal

VladislavsPerkanuks:


parse/ast.go
```diff
@@ -335,9 +330,10 @@ func (ae AnnotationExpression) validate() error {
 
 type QuotedLiteral string
 
-func (QuotedLiteral) node()    {}
-func (QuotedLiteral) literal() {}
-func (QuotedLiteral) value()   {}
+type NameLiteral string
+
+type NumberLiteral float64
```
klavinsjanis, line 335: shouldn't we support all numeric types?

klavinsjanis: Only allowed to pass in float64 as function argument?
What about supporting other types like int, unit, complex etc and approximations?

parse/ast.go
```diff
@@ -347,6 +343,8 @@ func (ql QuotedLiteral) String() string {
 
 	return fmt.Sprintf("|%s|", r.Replace(string(ql)))
 }
+func (nl NameLiteral) String() string   { return string(nl) }
+func (nl NumberLiteral) String() string { return fmt.Sprint(float64(nl)) }
```
mvilks, line 347: Why do you need additional `float64()` cast?

parse/ast.go
VladislavsPerkanuks: its not the scope of this PR, as it was not changed only moved.

but you can cast int, unit, exponent, to float64, with no problems.
The only problem is complex numbers

parse/ast.go
VladislavsPerkanuks: without casting infinite recursion :)

parse/ast.go
klavinsjanis: I guess add a comment :)

parse/ast.go
VladislavsPerkanuks: as fmt.Sprint(nl), will call nl.String(), which will cal fmt.Sprint(nl), and so on and on

mvilks: Indeed.

parse/ast.go
mvilks: complex numbers are not supported in MF2, the format is defined as JSON number https://www.rfc-editor.org/rfc/rfc8259#section-6

mvilks (approved):

parse/ast.go
VladislavsPerkanuks: yeah seems correct, so all other types are safely can be converted to float64, so no problemo

klavinsjanis (approved):

PR #46, opened 2024-02-01
Declarations: Full Implementation

VladislavsPerkanuks:


mvilks (approved):

parse/ast.go
```diff
@@ -440,7 +449,25 @@ func (ld LocalDeclaration) validate() error {
 	return nil
 }
 
-func (ReservedStatement) validate() error { return nil }
+func (rs ReservedStatement) validate() error {
+	if isZeroValue(rs.Keyword) {
```
klavinsjanis, line 453: could be good to add validation to check that the keyword is not  "input", "local", or ."match".

klavinsjanis (approved):

parse/ast.go
VladislavsPerkanuks: yup

PR #48, opened 2024-02-14
Initial MF2 Executer

VladislavsPerkanuks:
Current supported:
- Simple Message
- Expressions and Patterns

TODO:
- markup
- complex messages
- default function registry?
- locale ? [mf2 formating](https://github.com/unicode-org/message-format-wg/blob/122e64c2482b54b6eff4563120915e0f86de8e4d/spec/formatting.md#formatting-context)

> At a minimum, it includes:
>
>Information on the current locale, potentially including a fallback chain of locales. This will be passed on to formatting functions.


template/errors.go
```diff
@@ -0,0 +1,52 @@
...
+	ErrSyntax                = errors.New("syntax error")
+	ErrUnresolvedVariable    = errors.New("unresolved variable")
+	ErrUnknownFunction       = errors.New("unknown function reference")
+	ErrDuplicateOptionName   = errors.New("duplicate option name")
+	ErrUnsupportedExpression = errors.New("unsupported expression")
+	ErrFormatting            = errors.New("formatting error")
+)
+
+func syntaxErr(err error) error {
+	return errors.Join(ErrSyntax, err)
```
janishorsts, line 23: We MUST support golang 1.18. `errors.Join()` was added in v1.20 https://pkg.go.dev/errors#Join

VladislavsPerkanuks: i see, looking for altenative

template/errors.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"errors"
+	"fmt"
+
+	ast "go.expect.digital/mf2/parse"
+)
+
+// MessageFormat2 Errors as defined in the specification.
+//
+// https://github.com/unicode-org/message-format-wg/blob/122e64c2482b54b6eff4563120915e0f86de8e4d/spec/errors.md
```
janishorsts, line 12: 👏

template/template.go
```diff
@@ -0,0 +1,197 @@
...
+)
+
+type execFn func(operand any, opts map[string]any) (string, error)
+
+var execFuncs = map[string]execFn{}
+
+var mutex sync.Mutex
+
+// AddFunc adds a function to the template's function map.
+func AddFunc(name string, f func(any, map[string]any) (string, error)) {
```
janishorsts, line 19: *Avoid package-level variables in this repo - singletons.*

If you have a project importing two packages using the same function name but different implementations, the templating is broken.

Add functions to `type Template` instead.

template/template.go
```diff
@@ -0,0 +1,197 @@
...
+	case ast.ComplexMessage:
+		return fmt.Errorf("'%T' not implemented", message) // TODO: Implement.
+	}
+
+	s, err := resolve()
+	if err != nil {
+		return fmt.Errorf("resolve message: %w", err)
+	}
+
+	if _, err = wr.Write([]byte(s)); err != nil {
```
janishorsts, line 67: We should use `io.Writer` more efficiently. Pass it down to all functions that write string values and write to it directly.

template/errors.go
```diff
@@ -0,0 +1,43 @@
...
+func unsupportedExpressionErr(annotation ast.Annotation) error {
+	switch annotation.(type) {
+	default:
+		return ErrUnsupportedExpression
+	case ast.PrivateUseAnnotation:
+		return fmt.Errorf("%w with private use annotation: '%s'", ErrUnsupportedExpression, annotation)
+	case ast.ReservedAnnotation:
+		return fmt.Errorf("%w with reserved annotation: '%s'", ErrUnsupportedExpression, annotation)
+	}
+}
```
janishorsts, line 43: What is the benefit of these error functions? They are used only once and add boilerplate code.

```suggestion
```

template/template.go
```diff
@@ -0,0 +1,203 @@
...
+func Must(t *Template, err error) *Template {
+	if err != nil {
+		panic(err)
+	}
+
+	return t
+}
+
+// AddFunc adds a function to the template's function map.
+func (t *Template) AddFunc(name string, f execFn) {
```
janishorsts, line 50: Please use exposed types in the exposed function. Always.

?

```suggestion
func (t *Template) AddFunc(name string, f Func) {
```

template/template.go
```diff
@@ -0,0 +1,203 @@
...
+
+	return t, nil
+}
+
+func (t *Template) Execute(wr io.Writer, input map[string]any) error {
+	if t.ast == nil {
+		return errors.New("AST is nil")
+	}
+
+	t.executer = newExecuter(wr, input)
```
janishorsts, line 70: Would it make more sense to pass `io.Writer` to downstream functions instead to let the GC free memory when `io.Writer` is not needed anymore?

Current implementation does not release the memory when executor is reused multiple times.

template/template.go
VladislavsPerkanuks: yeah, makes sense

template/errors.go
VladislavsPerkanuks: yeah no benefits, you are right

template/template.go
VladislavsPerkanuks: Made an alternative, my previous implementation was kinda questionable.

template/template.go
```diff
@@ -0,0 +1,205 @@
...
+var (
+	ErrSyntax                = errors.New("syntax error")
+	ErrUnresolvedVariable    = errors.New("unresolved variable")
+	ErrUnknownFunction       = errors.New("unknown function reference")
+	ErrDuplicateOptionName   = errors.New("duplicate option name")
+	ErrUnsupportedExpression = errors.New("unsupported expression")
+	ErrFormatting            = errors.New("formatting error")
+)
+
+type ExecFn func(operand any, opts map[string]any) (string, error)
```
janishorsts, line 24: Why is it named `ExecFn`? Why not simply `Func`?

VladislavsPerkanuks: sure

template/template.go
```diff
@@ -0,0 +1,205 @@
...
+
+type Template struct {
+	ast       *ast.AST
+	execFuncs map[string]ExecFn
```
janishorsts, line 28: ?

```suggestion
	funcs map[string]Func
```

VladislavsPerkanuks: i was going to similarities with template pkg
https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/text/template/template.go;l=23

template/template.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// AddFunc adds a function to the template's function map.
+func (t *Template) AddFunc(name string, f ExecFn) {
+	t.execFuncs[name] = f
+}
+
+func (t *Template) Parse(input string) (*Template, error) {
```
janishorsts, line 36: Should we error on a function that is not declared in a Template but present in Message Format 2?

template/template.go
janishorsts, line 1: Please add comments to all exposed identifiers!

template/template.go
VladislavsPerkanuks: > To be conformant with MessageFormat 2.0, an implementation MUST implement all of the formatting and selection functions described in the default registry, including all of the options and option values, operands and outputs described by the default registry.

https://github.com/unicode-org/message-format-wg/blob/122e64c2482b54b6eff4563120915e0f86de8e4d/spec/registry.md#conformance-and-use

VladislavsPerkanuks: Out of scope of this PR, but we must add all Functions that are defined in 
https://github.com/unicode-org/message-format-wg/blob/122e64c2482b54b6eff4563120915e0f86de8e4d/spec/registry.xml

template/template.go
```diff
@@ -0,0 +1,205 @@
...
+	return &Template{execFuncs: make(map[string]ExecFn)}
+}
+
+type executer struct {
+	template *Template
+	wr       io.Writer
+	input    map[string]any
+}
+
+func (e *executer) writeString(s string) error {
```
janishorsts, line 78: KISS

```suggestion
func (e *executer) write(s string) error {
```

template/template.go
```diff
@@ -0,0 +1,205 @@
...
+func (e *executer) resolveExpression(expr ast.Expression) error {
+	value, err := e.resolveValue(expr.Operand)
+	if err != nil {
+		return fmt.Errorf("resolve value: %w", err)
+	}
+
+	if expr.Annotation == nil {
+		// NOTE: Parser won't allow value to be nil if annotation is nil.
+		valueStr := fmt.Sprint(value) // TODO: If value does not implement fmt.Stringer, what then ?
+		return e.writeString(valueStr)
```
janishorsts, line 127:
```suggestion
		s := fmt.Sprint(value) // TODO: If value does not implement fmt.Stringer, what then ?
		return e.writeString(s)
```

template/template.go
janishorsts: The answer is no. The functions can be added after parsing which is a valid use case.

```go
if t, err = t.Parse("..."); err != nil {
}

t.AddFunc("person", ...)

```

VladislavsPerkanuks: okay now i understand what you meant

template/example_test.go
VladislavsPerkanuks, line 1: https://go.dev/blog/examples

kaaposc (approved):

DarjaLeonova (approved):

template/example_test.go
```diff
@@ -0,0 +1,61 @@
...
+		if options == nil {
+			return strings.ToLower(fmt.Sprint(operand)), nil
+		}
+
+		if style, ok := options["style"].(string); ok && style == "first" {
+			return strings.ToLower(fmt.Sprint(operand)[0:1]) + fmt.Sprint(operand)[1:], nil
+		}
+
+		return "", fmt.Errorf("bad options")
+	})
```
janishorsts, line 36: `lower` does not retain behaviour symmetry to `upper`.

I suggest to use other function with options - e.g. `trim` https://cs.opensource.google/go/go/+/go1.22.0:src/strings/strings.go;l=891

template/template.go
```diff
@@ -0,0 +1,207 @@
...
+func (e *executer) execute() error {
+	switch message := e.template.ast.Message.(type) {
+	default:
+		return fmt.Errorf("unknown message type: '%T'", message)
+	case nil:
+		return nil
+	case ast.SimpleMessage:
+		return e.resolveSimpleMessage(message)
+	case ast.ComplexMessage:
+		return errors.New("complex message not implemented") // TODO: Implement.
```
janishorsts, line 98: I added the issue https://github.com/expect-digital/go-mf2/issues/50

template/template.go
```diff
@@ -0,0 +1,207 @@
...
+		case ast.TextPattern:
+			if err := e.write(string(pattern)); err != nil {
+				return err
+			}
+		case ast.Expression:
+			if err := e.resolveExpression(pattern); err != nil {
+				return fmt.Errorf("resolve expression: %w", err)
+			}
+		case ast.Markup: // TODO: Implement.
+			return fmt.Errorf("'%T' not implemented", pattern)
```
janishorsts, line 114: I added issue https://github.com/expect-digital/go-mf2/issues/51

template/example_test.go
VladislavsPerkanuks: This is just example, so it will be fine to leave as is

PR #53, opened 2024-03-01
Design And Initial Implementation of Default Function Registry

VladislavsPerkanuks:
Initial implementation design for integrating functions (default and custom) into the template:

Currently, only two functions have been included:

- `string`
- `number` (though not fully implemented due to complexity in accommodating all options

### DIFF
--- a/template/registry/number.go
+++ b/template/registry/number.go
@@ -1,0 +1,231 @@
+package registry
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+
+	"golang.org/x/text/currency"
+)
+
+// https://github.com/unicode-org/message-format-wg/blob/122e64c2482b54b6eff4563120915e0f86de8e4d/spec/registry.xml#L147
+
+var numberRegistryF = &Func{
+	Name:           "number",
+	Description:    "Locale-sensitive number formatting",
+	F:              numberF,
+	MatchSignature: nil, // Not allowed to use in matching context
+	FormatSignature: &Signature{
+		Input: true,
+		ValidateInput: func(a any) error {
+			if _, err := castAs[float64](a); err != nil {
+				return fmt.Errorf("unsupported type: %T: %w", a, err)
+			}
+
+			return nil
+		},
+		Options: Options{
+			{
+				Name:           "compactDisplay",
+				Description:    `Only used when notation is "compact".`,
+				PossibleValues: []any{"short", "long"},
+				Default:        "short",
+			},
+			{
+				Name: "currency",
+				Description: `The currency to use in currency formatting.
+Possible values are the ISO 4217 currency codes, such as "USD" for the US dollar,
+"EUR" for the euro, or "CNY" for the Chinese RMB — see the
+Current currency &amp; funds code list
+(https://www.unicode.org/cldr/charts/latest/supplemental/detailed_territory_currency_information.html).
+There is no default value; if the style is "currency", the currency property must be provided.`,
+				ValidateValue: func(a any) error {
+					_, err := castAs[currency.Unit](a)
+					return err
+				},
+			},
+			{
+				Name:           "currencyDisplay",
+				Description:    `How to display the currency in currency formatting.`,
+				PossibleValues: []any{"code", "symbol", "narrowSymbol", "name"},
+			},
+			{
+				Name: "currencySign",
+				Description: `In many locales, accounting format means to wrap the number with parentheses
+instead of appending a minus sign. You can enable this formatting by setting the
+currencySign option to "accounting".`,
+				PossibleValues: []any{"standard", "accounting"},
+				Default:        "standard",
+			},
+			{
+				Name:           "notation",
+				Description:    "The formatting that should be displayed for the number.",
+				PossibleValues: []any{"standard", "scientific", "engineering", "compact"},
+				Default:        "standard",
+			},
+			{
+				Name:        "numberingSystem",
+				Description: "Numbering system to use.",
+				PossibleValues: []any{
+					"arab", "arabext", "bali", "beng", "deva", "fullwide", "gujr", "guru",
+					"hanidec", "khmr", "knda", "laoo", "latn", "limb", "mlym", "mong", "mymr", "orya", "tamldec",
+					"telu", "thai", "tibt",
+				},
+			},
+			{
+				Name:           "signDisplay",
+				Description:    `When to display the sign for the number. "negative" value is Experimental.`,
+				PossibleValues: []any{"auto", "always", "exceptZero", "negative", "never"},
+				Default:        "auto",
+			},
+			{
+				Name:           "style",
+				Description:    "The formatting style to use.",
+				PossibleValues: []any{"decimal", "currency", "percent", "unit"},
+				Default:        "decimal",
+			},
+			{
+				Name: "unit",
+				Description: `The unit to use in unit formatting.
+Possible values are core unit identifiers, defined in UTS #35, Part 2, Section 6.
+A subset of units from the full list was selected for use in ECMAScript.
+Pairs of simple units can be concatenated with "-per-" to make a compound unit.
+There is no default value; if the style is "unit", the unit property must be provided.`,
+				ValidateValue: isPositiveInteger,
+			},
+			{
+				Name:           "unitDisplay",
+				Description:    "The unit formatting style to use in unit formatting.",
+				PossibleValues: []any{"long", "short", "narrow"},
+				Default:        "short",
+			},
+			{
+				Name: "minimumIntegerDigits",
+				Description: `The minimum number of integer digits to use.
+A value with a smaller number of integer digits than this number will be
+left-padded with zeros (to the specified length) when formatted.`,
+				ValidateValue: isPositiveInteger,
+				Default:       1,
+			},
+			{
+				Name: "minimumFractionDigits",
+				Description: `The minimum number of fraction digits to use.
+The default for plain number and percent formatting is 0;
+the default for currency formatting is the number of minor unit digits provided by
+the ISO 4217 currency code list (2 if the list doesn't provide that information).`,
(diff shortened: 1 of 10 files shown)


template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
+package registry
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+type Registry []Func
```
janishorsts, line 9: Why not `map[string]Func`? It speeds up Func lookup.

template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, fn := range r {
+		if fn.Name == name {
+			return fn, true
+		}
+	}
+
+	return Func{}, false
+}
+
+func (r *Func) Format(input any, options map[string]any) (any, error) {
```
janishorsts, line 56: ? 

```suggestion
func (f *Func) Format(input any, options map[string]any) (any, error) {
```

template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	if err := r.FormatSignature.check(input, options); err != nil {
+		return "", fmt.Errorf("check input: %w", err)
+	}
+
+	return r.F(input, options)
+}
+
+func (r *Func) Match(input any, options map[string]any) (any, error) {
```
janishorsts, line 68: ?

```suggestion
func (f *Func) Match(input any, options map[string]any) (any, error) {
```

template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := r.MatchSignature.check(input, options); err != nil {
+		return "", fmt.Errorf("check input: %w", err)
+	}
+
+	return r.F(input, options)
+}
+
+func (s *Signature) check(input any, options map[string]any) error {
+	if s.Input && input == nil {
+		return errors.New("input is nil, but required")
```
janishorsts, line 82: Alternatively

```suggestion
		return errors.New("input is required, got nil")
```

template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	if !s.Input && input != nil {
+		return errors.New("input is not allowed")
+	}
+
+	if s.Options == nil && len(options) > 0 {
```
janishorsts, line 89: Is it supposed to allow an empty slice in `s.Options`?

```suggestion
	if len(s.Options) == 0 && len(options) > 0 {
```

VladislavsPerkanuks: nope, u right

template/registry/registry.go
VladislavsPerkanuks: dunno :)

template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
...
+func (o Options) check(got map[string]any) error {
+	for name, val := range got {
+		opt, ok := o.find(name)
+		if !ok {
+			validOptions := make([]string, 0, len(o))
+			for _, opt := range o {
+				validOptions = append(validOptions, opt.Name)
+			}
+
+			return fmt.Errorf("unknown option: '%s'. Valid options: [%s]", name, strings.Join(validOptions, ", "))
```
janishorsts, line 117: Alternatively

```suggestion
			return fmt.Errorf("expected one of options - %s, got %s", name, strings.Join(validOptions, ", "))
```

template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		if opt.ValidateValue != nil {
+			if err := opt.ValidateValue(val); err != nil {
+				return fmt.Errorf("validate options value '%s': %w", name, err)
```
janishorsts, line 122:
```suggestion
				return fmt.Errorf("validate option value '%s': %w", name, err)
```

template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
...
+				if possible == val {
+					found = true
+					break
+				}
+			}
+
+			if !found {
+				return fmt.Errorf("invalid value '%s' for option '%s'. Valid values: %s  ", val, name, opt.PossibleValues)
+			}
+		}
```
janishorsts, line 139: Does our golang version support `slices.Contains` in this project?

```suggestion
		if !slices.Contains(opt.PossibleValues, val) {
			return fmt.Errorf("expected one of option '%s' allowed values - %s, got %s ", name, opt.PossibleValues, val)
		}
```

VladislavsPerkanuks: did not want to install additonal pkg, for one time :)

template/registry/registry.go
```diff
@@ -0,0 +1,153 @@
...
+	"strings"
+)
+
+type Registry []Func
+
+// Func is a function that can be used in formatting and matching contexts.
+type Func struct {
+	FormatSignature *Signature                             // Signature of the function when called in formatting context
+	MatchSignature  *Signature                             // Signature of the function when called in matching context
+	F               func(any, map[string]any) (any, error) // Function itself
```
janishorsts, line 15: `Func` or `Fn` instead of `F`?

template/registry/number.go
```diff
@@ -0,0 +1,231 @@
...
+			{
+				Name: "currency",
+				Description: `The currency to use in currency formatting.
+Possible values are the ISO 4217 currency codes, such as "USD" for the US dollar,
+"EUR" for the euro, or "CNY" for the Chinese RMB — see the
+Current currency &amp; funds code list
+(https://www.unicode.org/cldr/charts/latest/supplemental/detailed_territory_currency_information.html).
+There is no default value; if the style is "currency", the currency property must be provided.`,
+				ValidateValue: func(a any) error {
+					_, err := castAs[currency.Unit](a)
```
janishorsts, line 43: This does not validate currency value, right?

VladislavsPerkanuks: in this context: 
only `currency.Unit` can be passed as option value

VladislavsPerkanuks: so no casting

VladislavsPerkanuks: i mean, there is nothing to validate, if argument is currency.Unit, no ? 🤔

VladislavsPerkanuks: i think alternative is better
```go
if _, ok := a.(currency.Unit); !ok {
return ...
}

return nil
```

janishorsts: Mostly it seems to be ok, `currency.Unit` does not expose its value -  https://cs.opensource.google/go/x/text/+/refs/tags/v0.14.0:currency/currency.go;l=69

Effectively, consumers of the currency package cannot change the index directly.

However, zero value `currency.Unit` might cause an issue.

VladislavsPerkanuks: yeah, it can be a problem

template/registry/registry.go
VladislavsPerkanuks: wont work 

> any to satisfy comparable requires go1.20 or later

template/registry/registry.go
```diff
@@ -0,0 +1,146 @@
...
+	"strings"
+)
+
+type Registry map[string]Func
+
+// Func is a function that can be used in formatting and matching contexts.
+type Func struct {
+	FormatSignature *Signature                             // Signature of the function when called in formatting context
+	MatchSignature  *Signature                             // Signature of the function when called in matching context
+	Fn              func(any, map[string]any) (any, error) // Function itself
```
kaaposc, line 15: Is there a case when first returned value is other than string?

VladislavsPerkanuks: it could be anything :), thats the problem.

kaaposc (approved):

template/registry/registry.go
```diff
@@ -0,0 +1,146 @@
...
+	Default     any
+
+	ValidateValue  func(any) error // If option value is not restricted by a set of values.
+	PossibleValues []any           // If option value is restricted by a set of values.
+}
+
+type Options []Option
+
+// NewRegistry returns a new registry with default functions.
+func NewRegistry() Registry {
```
janishorsts, line 40: Consumers use `registry.New()`. This package unlikely to have other `New` functions.

```suggestion
func New() Registry {
```

template/registry/number.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// TODO: supports only style and signDisplay options.
+func numberF(input any, options map[string]any) (any, error) {
+	num, err := castAs[float64](input)
+	if err != nil {
+		return nil, fmt.Errorf("convert input to float64: %w", err)
+	}
+
+	if options == nil {
```
janishorsts, line 161: Check `len` instead?

```suggestion
	if len(options) == 0 {
```

template/registry/number.go
```diff
@@ -0,0 +1,240 @@
...
+	style, ok := options["style"]
+	if !ok {
+		style = "decimal"
+	}
+
+	switch style {
+	case "decimal":
+		result = strconv.FormatFloat(num, 'f', -1, 64)
+	case "percent":
+		result = fmt.Sprintf("%.2f%%", num*100) //nolint:gomnd
```
janishorsts, line 185: Does the spec imply to use 2 decimal points by default?

VladislavsPerkanuks: for percentage yes

template/registry/registry.go
```diff
@@ -0,0 +1,146 @@
...
+			return fmt.Errorf("validate input: %w", err)
+		}
+	}
+
+	if len(options) > 0 {
+		if err := s.Options.check(options); err != nil {
+			return fmt.Errorf("check options: %w", err)
+		}
+	}
+
```
janishorsts, line 95: Let the `check()` handle empty options: keeps the flow nice and tidy.

```suggestion
	if err := s.Options.check(options); err != nil {
		return fmt.Errorf("check options: %w", err)
	}
```

janishorsts: `check` already handles empty options nicely

PR #55, opened 2024-03-12
50-execute-complex-message-pt-2

DarjaLeonova:
Currently, in algorithms, I use the same variable names as in documentation, should we rename them to more readable ones?
https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#pattern-selection


template/template.go
```diff
@@ -296,3 +312,196 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+func (e *executer) resolveSelector() ([]Selector, error) {
+	// Step 1: Resolve Selector Value (Modified)
+	var res []Selector
+
+	for k, v := range e.input {
+		rv := Selector{k, v}
+		if e.input[k] != "" {
+			res = append(res, rv)
+		} else {
+			return nil, fmt.Errorf("%w '%s'", ErrSelection, res)
```
kaaposc, line 355: Did you mean to print out the `rv`?

kaaposc (approved):

template/template.go
```diff
@@ -296,3 +312,196 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+func (e *executer) resolvePreferences(m ast.Matcher, res []Selector) [][]string {
+	// Step 2: Resolve Preferences
+	var pref [][]string //nolint:prealloc
+
+	for i := range res {
+		var keys []string
+
+		for _, variant := range m.Variants {
+			for _, vKey := range variant.Keys {
+				key := vKey
```
janishorsts, line 371: Why copy the value twice (before go1.22)? In `range` and on this line?

template/template.go
```diff
@@ -296,3 +312,196 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+			sortable[tupleIndex].Score = currentScore
+		}
+
+		SortVariants(sortable)
+	}
+
+	return sortable
+}
+
+func (e *executer) selectBestVariant(sortable []SortableVariant) (strings.Builder, error) { //nolint:unparam
```
janishorsts, line 446: Why do you return `strings.Builder` instead of simple `string`?

template/template.go
```diff
@@ -296,3 +312,196 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+	}
+
+	return output, nil
+}
+
+func (e *executer) resolveSelector() ([]Selector, error) {
+	// Step 1: Resolve Selector Value (Modified)
+	var res []Selector
+
+	for k, v := range e.input {
```
VladislavsPerkanuks, line 350: > For each selector sel, in source order

so the loop should be across matcher expressions, not the input map. The whole idea seems wrong to me :)
Here is my draft, which should be correct

```go
func (e *executer) resolveSelector2(matcher ast.Matcher) ([]any, error) {
	res := make([]any, 0, len(matcher.MatchStatements))

	for _, selectExpr := range matcher.MatchStatements {
		var function ast.Function

		switch annotation := selectExpr.Annotation.(type) {
		case nil:
			// error: Select Expression cannot be without annotation
		case ast.ReservedAnnotation, ast.PrivateUseAnnotation:
			// error: Select Expression cannot be with reserved or private use annotation
		case ast.Function:
			function = annotation
		}

		registryF, ok := e.template.funcRegistry[function.Identifier.Name]
		if !ok {
			// error: Unknown function reference
		}

		opts, err := e.resolveOptions(function.Options)
		if err != nil {
			// error: Resolve Options
		}

		rslt, err := registryF.Match(selectExpr.Operand, opts)
		if err != nil {
			// error: Selection Error
		}

		res = append(res, rslt)

	}

	return res, nil
}
```

template/template_test.go
```diff
@@ -150,6 +150,55 @@ func Test_ExecuteComplexMessage(t *testing.T) {
 			},
 			expected: "Click here standalone ",
 		},
+		{
+			name: "matcher many",
+			args: args{
+				inputStr: ".match { $n } 0 {{no apples}} 1 {{{ $n } apple}} * {{{ $n } apples}}",
```
VladislavsPerkanuks, line 156: error, no select function

template/template.go
```diff
@@ -296,3 +312,196 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+
+	return matches
+}
+
+type SortableVariant struct {
+	Variant ast.Variant
+	Score   int
+}
+
+type SortableVariantSlice []SortableVariant
```
janishorsts, line 485: Avoid using type name in identifier names unless the context requires it.

Insteaf of `Slice`, you can use plural name.

```suggestion
type SortableVariants []SortableVariant
```

template/template_test.go
```diff
@@ -150,6 +150,55 @@ func Test_ExecuteComplexMessage(t *testing.T) {
 			},
 			expected: "Click here standalone ",
 		},
+		{
+			name: "matcher many",
```
VladislavsPerkanuks, line 154: i would write new test function for mathcer, to reduce repetition, when we testing different match cases
something like that
```go
func Test_Matcher(t *testing.T) {
	t.Parallel()

	tests := []struct {
		name      string
		inputStr  string
		inputMaps []map[string]any
		expected  []string
		funcs     []registry.Func
	}{
		{
			name:     "matcher string",
			inputStr: `.match { $n :string } no {{no apples}} one {{{ $n } apple}} * {{{ $n } apples}}`,
			inputMaps: []map[string]any{
				{"n": "no"},
				{"n": "one"},
				{"n": "many"},
			},
			expected: []string{"no apples", "one apple", "many apples"},
		},
	}

	for _, tt := range tests {
		if len(tt.inputMaps) != len(tt.expected) {
			t.Error("Arguments and expected results should have the same length")
		}

		tt := tt
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()

			template, err := New().Parse(tt.inputStr)
			require.NoError(t, err)

			for i, inputMap := range tt.inputMaps {
				actual, err := template.Sprint(inputMap)
				require.NoError(t, err)

				require.Equal(t, tt.expected[i], actual)
			}
		})
	}
}
```

template/template_test.go
```diff
@@ -266,6 +315,46 @@ func Test_ExecuteErrors(t *testing.T) {
...
+			expectedExecuteErr: ErrUnknownFunction,
+		},
+		{
+			name: "Plural Format Selection",
+			args: args{
+				inputStr: ".match {$count :string param=1 param=2} one {{Category match}} 1 {{Exact match}} *   {{Other match}}",
+				inputMap: map[string]any{"sda": "1"},
+			},
+			expectedExecuteErr: ErrDuplicateOptionName,
+		},
```
VladislavsPerkanuks, line 357: these have been already tested
```suggestion
```

template/template.go
```diff
@@ -296,3 +308,238 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
 
 	return m, nil
 }
+
+func (e *executer) patternSelection(m ast.Matcher) (string, error) {
```
VladislavsPerkanuks, line 312: resolveMatcher, would be more appropriate i guess

template/template.go
```diff
@@ -296,3 +308,238 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+	res, err := e.resolveSelector(m)
+	if err != nil {
+		return "", err
```
VladislavsPerkanuks, line 315: wrapping would be nice

template/template.go
```diff
@@ -296,3 +308,238 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+func (e *executer) resolvePreferences(m ast.Matcher, res []any) [][]string {
+	// Step 2: Resolve Preferences
+	pref := make([][]string, 0, len(res))
+
+	for i := range res {
+		var keys []string
+
+		for _, variant := range m.Variants {
+			for _, vKey := range variant.Keys {
+				if vKey.String() != "*" {
```
VladislavsPerkanuks, line 382: redundant if, already covered by switch statement

template/template.go
```diff
@@ -296,3 +308,238 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+func (e *executer) filterVariants(m ast.Matcher, pref [][]string) []ast.Variant {
+	// Step 3: Filter Variants
+	var filteredVariants []ast.Variant
+
+	for _, variant := range m.Variants {
+		matchesAllSelectors := true
+
+		for i, keyOrder := range pref {
+			key := variant.Keys[i]
+			if key.String() == "*" {
```
VladislavsPerkanuks, line 415: redundant if, already covered by switch statement

template/template.go
```diff
@@ -296,3 +308,238 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+	}
+
+	for i := len(pref) - 1; i >= 0; i-- {
+		matches := pref[i]
+
+		for tupleIndex, tuple := range sortable {
+			key := tuple.Variant.Keys[i]
+			currentScore := len(matches)
+
+			if key.String() != "*" {
```
VladislavsPerkanuks, line 461: redundant if, already covered by switch statement

template/template.go
```diff
@@ -296,3 +308,238 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+					ks = string(key)
+				case ast.QuotedLiteral:
+					ks = string(key)
+				case ast.NumberLiteral:
+					ks = fmt.Sprint(key)
+				}
+
+				if position := findPosition(ks, matches); position != -1 {
+					currentScore = position
+				}
```
VladislavsPerkanuks, line 477: is it crucial that score can be negative ? 🤔 
```suggestion
currentScore :=  findPosition(ks, matches)
```

template/template.go
```diff
@@ -296,3 +308,238 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+			sortable[tupleIndex].Score = currentScore
+		}
+
+		sort.Sort(SortableVariants(sortable))
+	}
+
+	return sortable
+}
+
+func (e *executer) selectBestVariant(sortable []SortableVariant) (string, error) { //nolint:unparam
```
VladislavsPerkanuks, line 489: ?

so this function always returns empty string, what the point of that ? 🤔 
that makes function patternSelection also always return empty string

template/template.go
```diff
@@ -296,3 +308,238 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+
+func (s SortableVariants) Less(i, j int) bool {
+	return s[i].Score < s[j].Score
+}
+
+func (s SortableVariants) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func findPosition(ks string, matches []string) int {
```
VladislavsPerkanuks, line 537: as you imported slices package, you could use `slices.Index ` instead of this func

template/template_test.go
```diff
@@ -150,6 +150,31 @@ func Test_ExecuteComplexMessage(t *testing.T) {
...
+			expected: "Foo and bar",
+		},
+		{
+			name: "Plural Format Selection",
+			args: args{
+				inputStr: ".match {$count :string} one {{Category match}} 1 {{Exact match}} *   {{Other match}}",
+				inputMap: map[string]any{"count": "1"},
+			},
+			expected: "Exact match",
+		},
```
VladislavsPerkanuks, line 177: matcher has its own tests now, so remove or move

template/template_test.go
```diff
@@ -172,6 +172,71 @@ func Test_ExecuteComplexMessage(t *testing.T) {
 	}
 }
 
+func Test_Matcher(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name      string
+		inputStr  string
+		inputMaps []map[string]any
```
janishorsts, line 181: Alternative names for `inputStr`: `text`, `msg`, `message`.

```suggestion
		text  string
		inputs []map[string]any
```

template/template.go
```diff
@@ -296,3 +303,218 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+
+		for _, variant := range m.Variants {
+			for _, vKey := range variant.Keys {
+				switch key := vKey.(type) {
+				case ast.CatchAllKey:
+					continue
+				case ast.NameLiteral:
+					keys = append(keys, string(key))
+				case ast.QuotedLiteral:
+					keys = append(keys, string(key))
```
janishorsts, line 383:
```suggestion
				case ast.NameLiteral, ast.QuotedLiteral:
					keys = append(keys, string(key))
```

In practice, we could use all three with `fmt.Sprint` but there is potentially a small performance penalty.

template/template.go
```diff
@@ -296,3 +303,218 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+
+			var ks string
+
+			switch key := key.(type) {
+			case ast.CatchAllKey:
+				continue
+			case ast.NameLiteral:
+				ks = string(key)
+			case ast.QuotedLiteral:
+				ks = string(key)
```
janishorsts, line 417:
```suggestion
			case ast.NameLiteral, ast.QuotedLiteral:
				ks = string(key)
```

template/template.go
```diff
@@ -296,3 +303,218 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+			var ks string
+
+			switch key := key.(type) {
+			case ast.CatchAllKey:
+				sortable[tupleIndex].Score = currentScore
+				continue
+			case ast.NameLiteral:
+				ks = string(key)
+			case ast.QuotedLiteral:
+				ks = string(key)
```
janishorsts, line 460:
```suggestion
			case ast.NameLiteral, ast.QuotedLiteral:
				ks = string(key)
```

template/template.go
```diff
@@ -296,3 +303,218 @@ func (e *executer) resolveOptions(options []ast.Option) (map[string]any, error)
...
+func matchSelectorKeys(rv any, keys []string) []string {
+	var matches []string
+
+	for _, key := range keys {
+		if value, ok := rv.(string); ok {
+			if key == value {
+				matches = append(matches, key)
+			}
+		}
+	}
```
janishorsts, line 498: We can simplify.

```suggestion
	value, ok := rv.(string)
	if !ok {
		return nil
	}
	
	var matches []string

	for _, key := range keys {
		if key == value {
			matches = append(matches, key)
		}
	}
```

template/template_test.go
```diff
@@ -172,6 +172,71 @@ func Test_ExecuteComplexMessage(t *testing.T) {
...
+			t.Parallel()
+
+			template, err := New().Parse(tt.inputStr)
+			require.NoError(t, err)
+
+			for i, inputMap := range tt.inputMaps {
+				actual, err := template.Sprint(inputMap)
+				require.NoError(t, err)
+
+				require.Equal(t, tt.expected[i], actual)
```
janishorsts, line 234: Nested table testing.

```suggestion
				t.Run(tt.expected[i], func (t *testing.T) {
					t.Parallel()

					actual, err := template.Sprint(inputMap)
					require.NoError(t, err)

					require.Equal(t, tt.expected[i], actual)
				})
```

janishorsts (approved):

template/template.go
VladislavsPerkanuks: we could not, Stringer for qutoed LIteral, prints string enclosed in |.
`|qutoed|`

template/template.go
VladislavsPerkanuks: is it allowed ?
If i remember correctly, type switch, with multiple types, then type is any, e.g. in this case key is any.
And you cannot cast any to string

DarjaLeonova: @VladislavsPerkanuks I resolved it by using a key.String(). I know that there is a behavior difference between string(key) and key.String(), but for now tests pass. Any objections?

VladislavsPerkanuks: I will close eyes 🙈

janishorsts: The underlying type of NameLiteral and QuotedLiteral is string. Always. Right?

It should not be possible to have a different type.

@VladislavsPerkanuks, under what circumstances can it be a different type?

VladislavsPerkanuks: they are both string yes, but this is not possible anyways
```go
case ast.NameLiteral, ast.QuotedLiteral:
  ks = string(key) // cannot convert key (variable of type parse.VariantKey) to type string
```

in type switch, when there is multiple types in one case, it fallbacks to the type its being switched to, in this case VariantKey

VladislavsPerkanuks: which is interface

PR #80, opened 2024-04-11
localise number option "style"

janishorsts:


template/registry/number.go
```diff
@@ -185,7 +185,8 @@ func numberFunc(input any, options map[string]any, locale language.Tag) (any, er
 	case "decimal":
 		result = p.Sprint(number.Decimal(num))
 	case "percent":
-		result = fmt.Sprintf("%.2f%%", num*100) //nolint:gomnd
+		const arbitrary = 6
```
mvilks, line 188: Why 6?

janishorsts: The clue is in the name - it is arbitrary. :)

It is unlikely that somebody needs more than 6 fraction digits for per cent values.

janishorsts: Good spot. However `the default for percent formatting is the larger of minimumFractionDigits and 0`.

janishorsts: `number.Percent()` has max fraction digits 0 by default.

Float stores approximate value, too high max fraction digits may produce approximate value. E.g.

```yaml
expected: "12,7%"
actual  : "12,7000000000000002%"
```

I bumped it up to 10.

janishorsts: I re-read the definition. It means, by default, `percent` has 0 fraction digits. I updated the implementation accordingly.

PR #86, opened 2024-04-17
add maximumSignificantDigits option to number func

janishorsts:


template/registry/number.go
```diff
@@ -189,9 +189,7 @@ func parseNumberOptions(opts Options) (*numberOptions, error) {
 	}
 
 	// The maximum number of significant digits to use.
-	const maxSignificantDigits = 21
-
-	options.MaximumSignificantDigits, err = opts.GetPositiveInt("maximumSignificantDigits", maxSignificantDigits, nil)
+	options.MaximumSignificantDigits, err = opts.GetPositiveInt("maximumSignificantDigits", -1, nil)
```
mvilks, line 192: Shouldn't the fallback in "get **positive** integer" function be, like, positive? :)

template/registry/number_test.go
```diff
@@ -87,4 +90,12 @@ func Test_Number(t *testing.T) {
...
 		map[string]any{"style": "percent", "minimumIntegerDigits": 3}, language.AmericanEnglish)
 	assert(0.01, "001%")
+
+	assert = assertFormat(t, numberRegistryFunc,
+		map[string]any{
+			"style":                    "percent",
+			"maximumFractionDigits":    5,
+			"maximumSignificantDigits": 4,
+		}, language.AmericanEnglish)
+	assert(0.12345, "12.34%")
```
mvilks, line 100: Please use some bigger number to show that the resulting number gets rounded not truncated
```suggestion
	assert(0.12346, "12.35%")
```

janishorsts: I'd like to keep it as it is - to emphasize it does not always round up.

I have not figured out how rounding works in the `number` package.

mvilks (approved):

template/registry/number.go
janishorsts: Fixed, renamed to `GetInt` and updated the func signature.

Issue #60, opened 2024-04-04
Pass Message Format 2 WG defined tests


janishorsts:
`wg_test.go` reads tests from MF2 WG and runs them.

Currently, most of them fail.

janishorsts:
Closed by accident.

PR #92, opened 2024-04-19
#60 - pass "{|2006-01-02T15:04:06| :datetime dateStyle=long}" MF2 WG …

janishorsts:


template/registry_datetime.go
```diff
@@ -39,12 +39,19 @@ func parseDatetimeInput(input any) (time.Time, error) {
...
 
-	v, ok := input.(time.Time)
-	if !ok {
-		return time.Time{}, fmt.Errorf("unsupported type %T: %w", input, ErrOperandMismatch)
-	}
+	switch v := input.(type) {
+	default:
+		return time.Time{}, fmt.Errorf("unsupported datetime type %T: %w", input, ErrOperandMismatch)
+	case string:
+		t, err := time.Parse(time.RFC3339[:len(v)], v)
```
mvilks, line 46: Interesting approach with dynamic format layout :+1:

mvilks (approved):

template/registry_datetime.go
janishorsts: Quick and dirty. It does not conform fully to ISO 8601 formatting. E.g. `"2006-01-02T15:0` is not valid ISO 8601. I'll fix it.

janishorsts: Not planning to fix it now. However, I added the comment.

PR #99, opened 2024-06-02
upgrade to latest MF2 WG

janishorsts:


parse/lex.go
```diff
@@ -629,8 +629,8 @@ func lexReservedBody(l *lexer) stateFn {
 		case v == '\\': // Reserved escape
 			v = l.next()
```
mvilks, line 630: Attn: the `v` gets reassigned. Won't it break the whole `switch`? Earlier (line 250) in similar case you're defining a new variable.

mvilks (approved):

parse/lex.go
janishorsts: Good spot. Renamed to `next`.

PR #131, opened 2024-07-01
130 - allow trailing whitespace for complex messages

janishorsts:
Parser already supports trailing whitespaces in the complex message.

I added a unit test to assert this behaviour.

### DIFF
--- a/parse/parse_test.go
+++ b/parse/parse_test.go
@@ -375,6 +375,18 @@ func TestParseComplexMessage(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:  "trailing whitespace",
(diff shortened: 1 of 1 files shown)


mvilks (approved):

parse/parse_test.go
```diff
@@ -375,6 +375,16 @@ func TestParseComplexMessage(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:  "trailing whitespace",
```
janishorsts, line 379: Not needed? The test is duplicated https://github.com/unicode-org/message-format-wg/pull/812/files

janishorsts:
Duplicate test. When we update the mf2 submodule, test will be present.

PR #136, opened 2024-07-03
Pass MF2WG string test #02

mvilks:
Collecting non-critical errors, e.g., when input variable not defined, it generates an error `unresolved-variable` but also executes the template using some fallback value (nil).

### DIFF
--- a/mf2wg_test.go
+++ b/mf2wg_test.go
@@ -39,8 +39,6 @@ func init() {
 		"TestMF2WG/.message-format-wg/test/tests/functions/integer.json/hello_{4.2_:integer}",
 		"TestMF2WG/.message-format-wg/test/tests/functions/integer.json/hello_{-4.20_:integer}",
 
-		"TestMF2WG/.message-format-wg/test/tests/functions/string.json/.match_{$foo_:string}_1_{{one}}_*_{{other}}#02",
-
 		"TestMF2WG/.message-format-wg/test/tests/functions/time.json/.local_$d_=_{|2006-01-02T15:04:06|_:date}_{{{$d_:time}}}",
 		"TestMF2WG/.message-format-wg/test/tests/functions/time.json/{|2006-01-02T15:04:06|_:time}",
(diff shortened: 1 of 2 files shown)


template/template.go
```diff
@@ -112,6 +112,11 @@ type executer struct {
 	template  *Template
 	w         io.Writer
 	variables map[string]any
+	errs      []error
```
janishorsts, line 115: not needed?

.message-format-wg
janishorsts, line 1: changing MF2 submodule commit is scope creep

template/template.go
```diff
@@ -424,7 +431,11 @@ func (e *executer) resolveSelector(matcher ast.Matcher) ([]any, error) {
 		}
 
 		input, err := e.resolveValue(selectExpr.Operand)
-		if err != nil {
+		if errors.Is(err, mf2.ErrUnresolvedVariable) {
```
janishorsts, line 434: This does not align with my comment on the ticket - how to handle errors in selectors from the MF2 spec.

janishorsts: Should `resolveValue()` return `ErrUnresolvedVariable` instead? Resolving a variable can occur in different places in the MF2. I understand `ErrUnresolvedVariable` should be returned everywhere we try to resolve a variable.

template/template.go
mvilks: It is needed to collect possible errors from several selectors

.message-format-wg
mvilks: reverted

mf2wg_test.go
```diff
@@ -260,7 +258,7 @@ func assertErr(t *testing.T, want Errors, err error) {
 
 	wantErr := func(want error) {
 		if !errors.Is(err, want) {
-			t.Errorf("want '%s', got '%s'", want, err)
+			t.Errorf("want error %s, got %s", want, err)
```
janishorsts, line 261: where did single quotes go? :)

janishorsts (approved):

PR #160, opened 2024-07-11
Pass MF2WG number test

mvilks:
Lex number literal at the end of lexName()

### DIFF
--- a/parse/lex.go
+++ b/parse/lex.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"strings"
 	"unicode/utf8"
-
-	"go.expect.digital/mf2"
 )
 
 // eof is the end of file item.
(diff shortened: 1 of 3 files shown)


parse/lex.go
```diff
@@ -465,14 +469,20 @@ func lexName(l *lexer) stateFn {
 		r rune   // current rune
 	)
 
-	for r = l.next(); isName(r); r = l.next() {
+	for r = l.next(); isName(r) || r == '+'; r = l.next() {
```
janishorsts, line 472: is `+` not allowed in name?

janishorsts: If it is not... does it not conflict with lexName() intent?

mvilks: starting char is: ALPHA / "_" / %xC0--> up with some exclusions
name char is: starting char / DIGIT / "-" / "." / %xB7 and something above 0x300
`+` is 0x2B so doesn't fit in any of previous allowed ranges

janishorsts: Does it imply we have to rename `lexName` to `lexNameOrNumber`?

janishorsts: Also. If `s` contains `+` and is NOT `number`, it is NOT `name`. Additional check required.

janishorsts (approved):

PR #170, opened 2024-07-15
161 pass mf2wg number test hello |42| number minimumfractiondigits=|2|

mvilks:
Try to parse strings into Int or Float when needed in number functions

### DIFF
--- a/template/registry.go
+++ b/template/registry.go
@@ -65,7 +65,7 @@ func (o Options) GetInt(name string, fallback int, validate ...Validate[int]) (i
 	if s, ok := v.(string); ok {
 		var err error
 
-		v, err = strconv.ParseInt(s, 10, 32)
+		v, err = strconv.ParseInt(s, 10, 64)
 		if err != nil {
 			return 0, fmt.Errorf(`parse integer from string "%s": %w`, s, err)
 		}
--- a/template/registry.go
+++ b/template/registry.go
@@ -3,6 +3,7 @@ package template
 import (
 	"fmt"
 	"slices"
+	"strconv"
 
 	"golang.org/x/exp/constraints"
 	"golang.org/x/text/language"
@@ -48,6 +49,15 @@ func (o Options) GetInt(name string, fallback int, validate ...Validate[int]) (i
 	v, ok := o[name]
 	if !ok {
 		return fallback, nil
+	}
+
+	if s, ok := v.(string); ok {
+		var err error
+
+		v, err = strconv.ParseInt(s, 10, 32)
+		if err != nil {
+			return 0, fmt.Errorf("parse integer from string '%s': %w", s, err)
+		}
 	}
 
 	i, err := castAs[int](v)
--- a/template/registry.go
+++ b/template/registry.go
@@ -67,7 +67,7 @@ func (o Options) GetInt(name string, fallback int, validate ...Validate[int]) (i
 
 		v, err = strconv.ParseInt(s, 10, 32)
 		if err != nil {
-			return 0, fmt.Errorf("parse integer from string '%s': %w", s, err)
+			return 0, fmt.Errorf(`parse integer from string "%s": %w`, s, err)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


template/registry.go
```diff
@@ -61,6 +62,15 @@ func (o Options) GetInt(name string, fallback int, validate ...Validate[int]) (i
 		return fallback, nil
 	}
 
+	if s, ok := v.(string); ok {
+		var err error
+
+		v, err = strconv.ParseInt(s, 10, 32)
+		if err != nil {
+			return 0, fmt.Errorf("parse integer from string '%s': %w", s, err)
```
janishorsts, line 70: We use double quotes in error messages for string values. ALWAYS. No single quotes.

template/registry_number.go
```diff
@@ -21,11 +22,21 @@ func parseNumberOperand(operand any) (float64, error) {
 		return 0, fmt.Errorf(format+": %w", append(args, mf2.ErrBadOperand)...)
 	}
 
+	var (
+		v   float64
+		err error
+	)
```
janishorsts, line 28: Declare variables as close as possible to their consumption.

template/registry_number.go
```diff
@@ -21,11 +22,21 @@ func parseNumberOperand(operand any) (float64, error) {
...
+
 	if operand == nil {
 		return errorf("operand is required")
 	}
 
-	v, err := castAs[float64](operand)
+	if s, ok := operand.(string); ok {
```
janishorsts, line 34: does it make sense to use type switch here?

mvilks: It's possible to cast string to []byte, not sure about other types

janishorsts:
```go
switch v := operand.(type) {
default:
   // castAs
   // return err if errored
case nil:
  // return err
case string:
  // json.Unmarshal
  // return err if errored
}
```

template/registry.go
```diff
@@ -61,6 +62,15 @@ func (o Options) GetInt(name string, fallback int, validate ...Validate[int]) (i
 		return fallback, nil
 	}
 
+	if s, ok := v.(string); ok {
+		var err error
+
+		v, err = strconv.ParseInt(s, 10, 32)
```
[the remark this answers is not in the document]
mvilks: The function returns `int` which is 32 bits, so to avoid unnecessary type casting.

template/registry_number.go
```diff
@@ -22,23 +22,24 @@ func parseNumberOperand(operand any) (float64, error) {
...
-
 	var (
 		v   float64
 		err error
 	)
 
-	if s, ok := operand.(string); ok {
-		err = json.Unmarshal([]byte(s), &v)
-	} else {
+	switch t := operand.(type) {
```
janishorsts, line 30:
```suggestion
        var (
            number float64
        )
	switch v := operand.(type) {
```

template/registry_number.go
```diff
@@ -22,23 +22,24 @@ func parseNumberOperand(operand any) (float64, error) {
...
-		return errorf("unsupported operand type %T: %w", operand, err)
+		if err != nil {
+			return errorf("unsupported operand type %T: %w", operand, err)
+		}
+	case nil:
+		return errorf("operand is required")
+	case string:
+		err = json.Unmarshal([]byte(t), &v)
+		if err != nil {
+			return errorf(`bad operand "%s": %w`, operand, err)
```
janishorsts, line 41: ok, what is the full error message chain returned here? Stuttering?

`bad operand` "%s": `bad operand`?

template/registry_number.go
janishorsts: How is `t` and `v` better?

`v` is idiomatic in Golang to represent a value, and in the given scenario, its scope is limited to the switch case, and it represents a typed value. Easy to follow its consumption.

`number` says it all. It is a number.

template/registry.go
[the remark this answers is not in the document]
janishorsts: `int` is not 32-bit.

If it is compiled for 64-bit arch, it is 64-bit.
If it is compiled for 32-bit arch, it is 32-bit. 

I'm not planning to support 32-bit at this stage.

janishorsts (approved):

PR #173, opened 2024-07-15
Leave unresolved operand as unresolved

mvilks:


template/template.go
```diff
@@ -174,11 +176,11 @@ func (e *executer) resolveDeclarations(declarations []ast.Declaration) error {
 			m[d.Operand] = struct{}{}
 
 			resolved, err := e.resolveExpression(ast.Expression(d))
+			e.variables[string(d.Operand.(ast.Variable))] = resolved //nolint: forcetypeassert // Will always be a variable.
```
janishorsts, line 179: Please leave a comment. This flow is unusual and deviates from the idiomatic Golang.

janishorsts:
Could you let me know what implementations you thought could solve this problem?

mvilks:
It was clear that the expected error is being received but the expression resolves to empty string, so the only implementation was to set the unresolved expression to something even if there's a resolution error.

janishorsts:
> It was clear that the expected error is being received but the expression resolves to empty string, so the only implementation was to set the unresolved expression to something even if there's a resolution error.

What happens if we have an `:uppercase` function and `.local {$foo :number} {{bar {$foo :uppercase}}}`, then it prints `bar {$FOO}`, right?

mvilks:
I have no idea if it is even valid template, and what it should return..

janishorsts:
We need to handle the cases correctly when the variable is unresolved. I'll create a ticket later.

janishorsts (approved):

PR #181, opened 2024-07-17
Implement time function

mvilks:


template/registry_time.go
```diff
@@ -0,0 +1,98 @@
...
+	"golang.org/x/text/language"
+)
+
+// timeFunc is the implementation of the time function. Locale-sensitive time formatting.
+var timeRegistryFunc = RegistryFunc{
+	Format: timeFunc,
+}
+
+type timeOptions struct {
+	// (default is system default time zone or UTC)
```
janishorsts, line 16: This comment does not align with the default value. Is the same applicable to `:date` comment?

```suggestion
	// (default is UTC)
```

mvilks: `timezone` option is not even defined for either `:date` or `:time`, so I was just assuming it would be appropriate to add it anyway.

template/registry_time.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	opts, err := parseTimeOptions(options)
+	if err != nil {
+		return errorf("%w", err)
+	}
+
+	if opts == nil {
+		opts = &timeOptions{Style: "short"}
+	}
```
janishorsts, line 78: This should be handled by parseTimeOptions

template/registry.go
```diff
@@ -136,3 +138,29 @@ func castAs[T any](val any) (T, error) {
 
 	return v.Interface().(T), nil //nolint:forcetypeassert
 }
+
+// getTZ gets the timezone information from the registry function options.
+func getTZ(options map[string]any) (*time.Location, error) {
+	var err error
+
+	// default is system default time zone or UTC
```
janishorsts, line 146:
```suggestion
	// default is system UTC
```

template/registry.go
```diff
@@ -136,3 +138,29 @@ func castAs[T any](val any) (T, error) {
...
+		return timezone, fmt.Errorf("want timeZone as string or *time.Location, got %T", v)
+	case *time.Location:
+		timezone = tz
+	case string:
+		if timezone, err = time.LoadLocation(tz); err != nil {
+			return timezone, fmt.Errorf("load TZ data for %s: %w", tz, err)
+		}
+	}
+
+	return timezone, nil
```
janishorsts, line 165:
```suggestion
	v, ok := options["timeZone"]
	if !ok {
		return time.UTC, nil
	}

	switch tz := v.(type) {
	default:
		return nil, fmt.Errorf("want timeZone as string or *time.Location, got %T", v)
	case *time.Location:
		return tz, nil
	case string:
		timezone, err = time.LoadLocation(tz);
		if err != nil {
			return nil, fmt.Errorf("load TZ data for %s: %w", tz, err)
		}
		return timezone
	}
```

template/registry_time.go
janishorsts: The comment does not reflect the implementation. :)

In our implementation, we never fall back to system time. It ALWAYS defaults to UTC.

template/registry.go
```diff
@@ -136,3 +138,25 @@ func castAs[T any](val any) (T, error) {
...
+	switch tz := v.(type) {
+	default:
+		return nil, fmt.Errorf("want timeZone as string or *time.Location, got %T", v)
+	case *time.Location:
+		return tz, nil
+	case string:
+		timezone, err := time.LoadLocation(tz)
+		if err != nil {
+			return timezone, fmt.Errorf("load TZ data for %s: %w", tz, err)
+		}
```
janishorsts, line 158:
```suggestion
		if err != nil {
			return nil, fmt.Errorf("load TZ data for %s: %w", tz, err)
		}
```

janishorsts (approved):

PR #183, opened 2024-07-23
Get plural form of the number

mvilks:


template/registry.go
```diff
@@ -20,6 +21,13 @@ type RegistryFunc struct {
 
 type Registry map[string]RegistryFunc
 
+type FuncContext int
+
+const (
+	Format FuncContext = iota
+	Match
+)
```
janishorsts, line 29: Where are these names coming from? In MF2 spec, I see `Selection` and `Formatting`.

janishorsts: Ok, these names are used in our code base.

mvilks: It's coming from the `RegistryFunc` definition
```golang
type RegistryFunc struct {
	Match  func(input any, options Options, locale language.Tag) (output any, err error)
	Format func(input any, options Options, locale language.Tag) (output any, err error)
}
```
but I guess MF2 spec naming would work, too.

template/registry_integer.go
```diff
@@ -8,20 +8,23 @@ import (
 
 // integerRegistryFunc is the implementation of the integer function. Locale-sensitive integer formatting.
 var integerRegistryFunc = RegistryFunc{
-	Format: integerFunc,
+	Format: integerFunc(Format),
```
janishorsts, line 11: Would changing RegistryFunc simplify?

> Functions are used to evaluate, format, select, or otherwise process data values during formatting.

> A function's entry in the function registry will define whether the function is a selector or formatter (or both)...

What do you think?

```go
type FuncKind int

const (
	Formatter FuncKind = iota
	Selector
)

type RegistryFunc = func(operand any, options Options, locale language.Tag, kind FuncKind) (any, error)
```

mvilks: That would change the way currently supported functions are determined.

janishorsts: I see. The current way is better for devex - e.g. custom functions don't need to handle kind.

Can we rename Match to Select to align naming with MF2 documentation?

template/registry.go
```diff
@@ -160,3 +168,21 @@ func getTZ(options map[string]any) (*time.Location, error) {
 		return timezone, nil
 	}
 }
+
+// toString formats plural.Form as string.
+func toString(f plural.Form) string {
```
janishorsts, line 173:
```suggestion
// pluralFormString formats plural.Form as string.
func pluralFormString(f plural.Form) string {
```

janishorsts (approved):

PR #187, opened 2024-07-29
184 execute expressions on demand

mvilks:


template/template.go
```diff
@@ -157,34 +181,37 @@ func (e *executer) resolveComplexMessage(message ast.ComplexMessage) error {
...
-			e.variables[string(d.Variable)] = resolved
-
-			if err != nil {
-				return fmt.Errorf("local declaration: %w", err)
-			}
-
+			e.variables[string(d.Variable)] = newVariable(d.Expression, nil)
 		case ast.InputDeclaration:
-			m[d.Operand] = struct{}{}
+			name := string(d.Operand.(ast.Variable)) //nolint: forcetypeassert // always ast.Variable
```
janishorsts, line 205: Can we use `ast.Variable` as key T in the map? To avoid converting `ast.Variable` to `string` all the time.

mvilks: Template params are saved in that map as `string: any`, this change would require converting strings into `ast.Variable`

template/template.go
```diff
@@ -157,34 +181,37 @@ func (e *executer) resolveComplexMessage(message ast.ComplexMessage) error {
...
-
-			if err != nil {
-				return fmt.Errorf("input declaration: %w", err)
+			val, ok := e.variables[name]
+			if !ok {
+				return mf2.ErrUnresolvedVariable
 			}
+
+			expr := ast.Expression(d)
+			expr.Operand = newLiteral(val)
```
janishorsts, line 213: If input contains the `time.Location` type, what happens? How do we handle this scenario?

janishorsts: We cannot store this value in any literal in the `parse` package.

mvilks: I guess that will cause some mild panic in some part of the template execution. In all other cases input is either string or number (float64 or int).

janishorsts: We require to support `time.Location` and others -  without serialization.

mvilks: Values from input eventually get used either as Operand for an Annotation or value for Option - in neither case `time.Location` is valid type. It most likely gets passed as a string (QuotedLiteral) and only the function (Annotation) interprets it as `time.Location`.

janishorsts: What MF2 expects in this case? is $tz input?

```go
tpl.Parse(`{{ |2024-08-01| :date timeZone={$tz} }}`)
tpl.Sprint(map[string]any{"tz": location.UTC})
```

mvilks: Yes, `tz` is input. In `resolveValue()` it's going to grab the value from variables and pass it into registry function Format() which will interpret it as needed. Looks like in this scenario the location type doesn't get serialized into string. I'll double check the MF2 spec docs.

janishorsts: #190 will probably solve it

janishorsts (approved):

PR #191, opened 2024-07-31
New ResolvedValue type

mvilks:


template/template.go
```diff
@@ -31,28 +31,57 @@ type Template struct {
 	locale   language.Tag
 }
 
-type variable struct {
-	name        *string
-	formatValue *string // TODO(mvilks): "selectValue" to be implemented
-	expression  ast.Expression
+type Result struct {
```
janishorsts, line 34: `Result` is very abstract and is missing intent. I suggest something like `Resolved`.

janishorsts: ... or `ResolvedValue`

template/template.go
```diff
@@ -31,28 +31,77 @@ type Template struct {
...
 
-	v.formatValue = &val
+func (r *Result) SelectKey(keys []string) string {
+	if r.selectKey != nil {
+		return r.selectKey(keys)
+	}
+
+	if r.format != nil {
+		return r.format()
+	}
```
janishorsts, line 78: What happens when `format()` value does not match any key in keys?

janishorsts: Also, if the struct has no format function, we can set a default function to select the key based on format value.

template/template.go
```diff
@@ -31,28 +31,77 @@ type Template struct {
...
+		return value.String()
+	case string, []byte, []rune, int, int8, int16, int32, int64,
+		uint, uint8, uint16, uint32, uint64, float32, float64, bool,
+		complex64, complex128, error:
+		return fmt.Sprint(value)
 	}
+}
 
-	v.formatValue = &val
+func (r *Result) SelectKey(keys []string) string {
```
janishorsts, line 71: Do mf2 package consumers need access to SelectKey?

template/template.go
janishorsts: The same applies to the format function. NewResult() sets default format() or selectKeys() if not set with `WithSelectKey` and `WithFormat` options.

janishorsts: However, what does MF2 say about using a registry function for selection and formatting?

mvilks: If matching any key fails, the fallback catch-all key is returned.

template/registry_number.go
```diff
@@ -321,6 +310,37 @@ func numberFunc(context FuncContext) func(operand any, options Options, locale l
...
+		form := plural.Cardinal.MatchDigits(locale, digits.Digits, int(digits.Exp), int(digits.End-digits.Exp))
+
+		return pluralFormString(form)
+	}
+
+	return NewResolvedValue(value, WithFormat(format), WithSelectKey(selectKey)), nil
+}
+
+func hasExactKey(keys []string) bool {
+	categories := []string{"zero", "one", "two", "few", "many", "other"}
```
janishorsts, line 337: Please use the switch for static lists. ALWAYS.

template/template.go
```diff
@@ -31,28 +31,83 @@ type Template struct {
...
 
-type variable struct {
-	name        *string
-	formatValue *string // TODO(mvilks): "selectValue" to be implemented
-	expression  ast.Expression
+type ResolvedValue struct {
+	value     any
+	selectKey func(keys []string) string
+	format    func() string
+	err       error
```
janishorsts, line 38: Please add comments where needed. Also, all exposed identifiers MUST have comments.

template/template.go
janishorsts: unicode-org/message-format-wg#852

janishorsts: How strict do we want to be about formatting?

* We MUST return an error anyway if no formatting function is present
* Fall back to fallback value or value. Currently, your implementation falls back to value.

template/template.go
```diff
@@ -31,28 +31,83 @@ type Template struct {
...
+
+	if v, ok := value.(*ResolvedValue); ok {
+		r = v
+	} else {
+		r = &ResolvedValue{
+			value:     value,
+			format:    func() string { return defaultFormat(value) },
+			selectKey: func(keys []string) string { return defaultSelectKey(value, keys) },
+		}
+	}
```
janishorsts, line 104:
```suggestion
	r, ok := value.(*ResolvedValue)
	if !ok {
		r = &ResolvedValue{
			value:     value,
			format:    func() string { return defaultFormat(value) },
			selectKey: func(keys []string) string { return defaultSelectKey(value, keys) },
		}
	}
```

template/template.go
```diff
@@ -200,18 +255,33 @@ func (e *executer) resolveDeclarations(declarations []ast.Declaration) error {
...
+			}
+
+			e.variables[string(d.Variable)] = r // newVariable(d.Expression, nil)
 		case ast.InputDeclaration:
 			name := string(d.Operand.(ast.Variable)) //nolint: forcetypeassert // always ast.Variable
 
 			val, ok := e.variables[name]
 			if !ok {
-				return mf2.ErrUnresolvedVariable
+				r := NewResolvedValue("{$" + name + "}")
```
janishorsts, line 269: Would it make sense to create a helper function?

```suggestion
				r := newFallbackValue(name)
```

template/template.go
```diff
@@ -200,18 +255,33 @@ func (e *executer) resolveDeclarations(declarations []ast.Declaration) error {
...
 			}
 
 			expr := ast.Expression(d)
 			expr.Operand = newLiteral(val)
-			e.variables[name] = newVariable(expr, &name)
+
+			r, err := e.resolveExpression(expr)
+			if err != nil {
+				r = NewResolvedValue("{$" + name + "}")
+				r.err = errors.Join(r.err, fmt.Errorf("resolve for %s: %w", name, err))
```
janishorsts, line 281: isn't r.err empty by default?
```suggestion
				r.err = fmt.Errorf("resolve for %s: %w", name, err)
```

template/template.go
```diff
@@ -249,10 +319,10 @@ func (e *executer) resolvePattern(pattern []ast.PatternPart) error {
 	return resolutionErr
 }
 
-func (e *executer) resolveExpression(expr ast.Expression) (string, error) {
+func (e *executer) resolveExpression(expr ast.Expression) (*ResolvedValue, error) {
```
janishorsts, line 322: Why does the consumer of `resolveExpression()` handle fallback value? Why doesn't `resolveExpression` return a fallback value?

janishorsts: https://github.com/translate-agent/mf2/pull/191/files#diff-d24cab31570f37b8fb2190891c0a53a42e2243d6d242df62c2ef59a7f9d8c972R278

mvilks: Because the expression sometimes has no information about the variable it was meant to bind to, e.g. in `.input` declarations.

template/template.go
```diff
@@ -415,6 +476,38 @@ func (e *executer) resolveMatcher(m ast.Matcher) error {
 	return matcherErr
 }
 
+func (e *executer) hasAnnotation(operand ast.Value) bool {
```
janishorsts, line 479: What happens in `.input {$n :number} .local $x = ${n} .match {$x} 1 {one} * {not one}`? `$x` has no annotation but it is a number.

janishorsts: You already do it. Just noticed recursive call.

template/template.go
```diff
@@ -431,6 +524,23 @@ func (e *executer) resolveSelector(matcher ast.Matcher) ([]any, error) {
 
 		switch annotation := selector.Annotation.(type) {
 		case nil:
+			if e.hasAnnotation(selector.Operand) {
```
janishorsts, line 527:
```suggestion
			if !e.hasAnnotation(selector.Operand) {
				return nil, mf2.ErrMissingSelectorAnnotation
			}
			
			// ...
```

template/template.go
janishorsts: `.input {$n :number} {{Count: $n}}`. What's missing in `{$n :number}`? Isn't fallback value `{$n}`?

mvilks: Currently the declaration resolving replaces the `$n` with the value from external input parameters to avoid circular reference problems. Maybe it's wise to keep declarations in a separate map from the external input params. I'll see if that solves problems.

mvilks: No need for a separate map, refactoring solved the problem.

template/template.go
```diff
@@ -31,28 +31,94 @@ type Template struct {
...
+func (r *ResolvedValue) String() string {
+	if r.format != nil {
+		return r.format()
+	}
+
+	return defaultFormat(r.value)
+}
+
+// OptionalFunction is a function to apply to the ResolvedValue.
+type OptionalFunction func(*ResolvedValue)
```
janishorsts, line 83: It is an option, not optional. Also, I'd like to suggest the following name to reflect its purpose.

```suggestion
type ResolvedValueOpt func(*ResolvedValue)
```

template/template.go
```diff
@@ -31,28 +31,94 @@ type Template struct {
...
+
+// FormatFunction is a format function signature.
+type FormatFunction func() string
+
+// SelectKeyFunction is a format function signature.
+type SelectKeyFunction func(keys []string) string
```
janishorsts, line 89: No need to define exposed functions. We can safely remove them.

template/template.go
```diff
@@ -31,28 +31,94 @@ type Template struct {
...
 	}
+}
 
-	val, err := e.resolveExpression(v.expression)
-	if err != nil && v.name != nil {
-		n := "{$" + *v.name + "}"
-		val = n
+// NewResolvedValue creates a new variable of type *ResolvedValue.
+// If value is already *ResolvedValue, the optional format() and selectKey() are applied to it.
+func NewResolvedValue(value any, functions ...OptionalFunction) *ResolvedValue {
```
janishorsts, line 107:
```suggestion
func NewResolvedValue(value any, opts ...ResolvedValueOpt) *ResolvedValue {
```

template/template.go
janishorsts: We'll resolve it in separate PR.

janishorsts (approved):

PR #196, opened 2024-08-06
Use *ResolvedValue in template variables

mvilks:


template/template.go
```diff
@@ -170,10 +170,35 @@ func (t *Template) Execute(w io.Writer, input map[string]any) error {
...
+		default:
+			executer.variables[k] = def
+			continue
+		case string:
+			function = "string"
+		case float64, int:
+			function = "number"
+		}
+
+		f, ok := t.registry[function]
```
janishorsts, line 190: Why look up a function every time we iterate through _input_ range? we can do it before _for_ loop

mvilks: Each input value can be a different type. Do you suggest iterating through the input slice to gather types and then again to create ResolvedValues?

janishorsts (approved):

template/template.go
janishorsts: We can access registry functions directly without looking them up.

mvilks: Using registry function directly.

PR #197, opened 2024-08-07
ResolvedValue in function input

mvilks:


template/template.go
```diff
@@ -76,6 +76,11 @@ func (r *ResolvedValue) String() string {
 	return defaultFormat(r.value)
 }
 
+// Value exposes raw value.
```
janishorsts, line 79: Alternatively

```suggestion
// Value returns raw value.
```

template/template.go
```diff
@@ -175,19 +180,19 @@ func (t *Template) Execute(w io.Writer, input map[string]any) error {
 	for k, v := range input {
 		var f Func
 
-		def := NewResolvedValue(v, WithFormat(func() string { return defaultFormat(v) }))
+		r := NewResolvedValue(v, WithFormat(func() string { return defaultFormat(v) }))
 
 		switch v.(type) {
 		default:
-			executer.variables[k] = def
+			executer.variables[k] = r
```
janishorsts, line 187:
```suggestion
		switch v.(type) {
		default:
			executer.variables[k] = NewResolvedValue(v, WithFormat(func() string { return defaultFormat(v) }))
```

template/example_test.go
```diff
@@ -113,3 +107,51 @@ func ExampleTemplate_complexMessage() {
 
 	// Output: John is 42 years old and his favorite color is 255,0,0.
 }
+
+func ExampleTemplate_calculator() {
```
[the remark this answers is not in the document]
mvilks: Partly to show that the raw value has to be exposed. But also it's cool imho ;)

template/template.go
mvilks: I'm using the `r` also as an input to the registry func on line 195.

template/example_test.go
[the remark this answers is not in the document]
janishorsts: Can you think of a better example?

The functions in the registry SHOULD be related to the localisation. I don't want to encourage developers to register functions that should be handled in Golang instead. Otherwise, it adds noise to the text and makes it harder for translators to maintain translations.

E.g. _:parity_ that returns the localised name for _odd_ or _even_.

The _:color_ example can be improved too.

template/template.go
janishorsts: Are you sure? :) It is a new declaration, not an assignment to the previous _r_ variable.

template/example_test.go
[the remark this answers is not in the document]
mvilks: Will add TODO and address it in another PR.

template/template.go
mvilks: Oh no, it bit me again..

template/example_test.go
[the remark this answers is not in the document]
janishorsts: Keep it simple, delete it. Less noise.

janishorsts (approved):

PR #204, opened 2024-08-08
203 pass mf2 wg syntax tests

mvilks:


mvilks:
Unknown-function errors handled.
Number literal formatting gives headache.

template/template.go
```diff
@@ -355,6 +355,10 @@ func (e *executer) resolveExpression(expr ast.Expression) (*ResolvedValue, error
 
 		switch v := expr.Operand.(type) {
 		default:
+			if f, ok := expr.Annotation.(ast.Function); ok {
+				return NewResolvedValue("{:" + f.Identifier.String() + "}")
```
janishorsts, line 359: What about a helper function _newFallbackValue_? We return fall-back value in many places.

mf2wg_test.go
```diff
@@ -255,14 +236,16 @@ func assertErr(t *testing.T, want Errors, err error) {
 			wantErr(mf2.ErrMissingFallbackVariant)
 		case "missing-selector-annotation":
 			wantErr(mf2.ErrMissingSelectorAnnotation)
+		case "syntax-error":
```
janishorsts, line 239: please ensure case are ORDERED by name

mvilks: yes, that's why I reordered them

template/template.go
mvilks: Implemented `newFallbackValue` that does all the type checking and correct wrapping.

template/template.go
```diff
@@ -109,6 +109,39 @@ func NewResolvedValue(value any, options ...ResolvedValueOpt) *ResolvedValue {
 	return r
 }
 
+func newFallbackValue(value any) *ResolvedValue {
```
janishorsts, line 112: I'd expect it to implement https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#fallback-resolution and nothing else.

janishorsts (approved):

PR #210, opened 2024-08-15
Handle duplicate variant errors

mvilks:
Search for duplicate variants before matching anything

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -465,6 +465,10 @@ func (e *executer) resolveMatcher(m ast.Matcher) error {
 		return fmt.Errorf("matcher: %w", matcherErr)
 	}
 
+	if hasDuplicateVariants(m.Variants) {
+		return fmt.Errorf("marcher: %w", mf2.ErrDuplicateVariant)
+	}
+
 	pref := e.resolvePreferences(m, res)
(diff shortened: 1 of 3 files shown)


template/template.go
```diff
@@ -696,6 +692,43 @@ func (e *executer) bestMatchedPattern(filteredVariants []ast.Variant, pref [][]s
 	return sortable[0].Variant.QuotedPattern
 }
 
+func keyString(key ast.VariantKey) string {
```
janishorsts, line 695: Can we use`String()`? All types have `String()`, right?

janishorsts: ok, it would return MF2 formatted value.

template/template.go
```diff
@@ -696,6 +692,43 @@ func (e *executer) bestMatchedPattern(filteredVariants []ast.Variant, pref [][]s
...
+	switch k := key.(type) {
+	case ast.CatchAllKey:
```
janishorsts, line 697:
```suggestion
	switch k := key.(type) {
	default:
		return ""
	case ast.CatchAllKey:
```

janishorsts (approved):
LGTM

PR #212, opened 2024-08-15
Additional whitespace checks

mvilks:


parse/lex.go
janishorsts, line 1: Lexer MUST NOT skip whitespace items.

parse/lex_test.go
```diff
@@ -536,6 +536,30 @@ func Test_lex(t *testing.T) {
...
 		},
+		{
+			name:  "head and tail whitespaces",
+			input: "  {{}}  ",
+			want: []item{
+				mk(itemWhitespace, "  "),
+				mk(itemQuotedPatternOpen, "{{"),
+				mk(itemQuotedPatternClose, "}}"),
+				mk(itemWhitespace, "  "),
+			},
```
janishorsts, line 547: I forgot to add EOF

```suggestion
				mk(itemWhitespace, "  "),
				mk(itemEOF, ""),
			},
```

mvilks: I wonder why it doesn't complain 🤔

mvilks:
```
goos: linux
goarch: amd64
pkg: go.expect.digital/mf2/parse
cpu: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
                        │ before.txt  │           after.txt           │
                        │   sec/op    │   sec/op     vs base          │
ComplexMessage_String-8   4.249µ ± 2%   4.293µ ± 2%  ~ (p=0.271 n=10)

                        │  before.txt  │            after.txt             │
                        │     B/op     │     B/op      vs base            │
ComplexMessage_String-8   20.33Ki ± 0%   20.33Ki ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

                        │ before.txt │           after.txt            │
                        │ allocs/op  │ allocs/op   vs base            │
ComplexMessage_String-8   43.00 ± 0%   43.00 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

parse/lex_test.go
```diff
@@ -550,9 +578,23 @@ func assertItems(t *testing.T, want []item, l *lexer) {
 
 	logItems := make([]func(), 0, len(want))
 
-	for _, wantItem := range want {
+	i := 0
```
janishorsts, line 581: Collecting items first would simplify the flow and make assertion more reliable.

parse/lex_test.go
```diff
@@ -550,21 +578,38 @@ func assertItems(t *testing.T, want []item, l *lexer) {
...
+
+		if itm.typ == itemError || itm.typ == itemEOF {
+			break
+		}
+	}
+
+	if len(got) != len(want) {
+		t.Errorf(`want %d items, got %d`, len(want), len(got))
+
+		return
```
janishorsts, line 595: If this fails, it is hard to understand why exactly - no visibility of items.

Can we simplify by using `slices.Equal()` now?

parse/lex.go
```diff
@@ -279,9 +296,13 @@ func lexPattern(l *lexer) stateFn {
 			}
 
 			return lexExpr(l)
-		case !l.isComplexMessage && len(s) == 0 && r == '.':
+		case !l.isComplexMessage && (len(s) == 0 || isStringWhitespace(s)) && r == '.':
```
janishorsts, line 299: `isStringWhitespace` is called twice - once here and second time in `mkText`. Should it be like this?

```suggestion
		case !l.isComplexMessage && isStringWhitespace(s) && r == '.':
			l.backup()
			return l.emitItem(mk(itemWhitespace, s))
		case !l.isComplexMessage && len(s) == 0 && r == '.':
```

PR #239, opened 2024-09-17
Mandatory whitespaces

mvilks:
Some of whitespaces in MF2 now are mandatory.

### DIFF
--- a/parse/parse.go
+++ b/parse/parse.go
@@ -643,8 +643,20 @@ selectorsLoop:
 		case itemEOF:
 			return errorf("%w", unexpectedErr(itm))
 		case itemVariable:
+			p.backup()
+			if v := p.current(); v.typ != itemWhitespace {
+				// there should be a whitespace before each selector
+				return errorf("%w", mf2.ErrSyntax)
+			}
+			p.next()
+
 			matcher.Selectors = append(matcher.Selectors, Variable(itm.val))
 		}
+	}
+
+	if v := p.current(); v.typ != itemWhitespace {
+		// there should be a whitespace between selectors and variants
+		return errorf("%w", mf2.ErrSyntax)
 	}
 
 	// parse one or more variants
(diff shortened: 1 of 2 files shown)


parse/parse.go
```diff
@@ -647,6 +652,11 @@ selectorsLoop:
 		}
 	}
 
+	if v := p.current(); v.typ != itemWhitespace {
+		// there should be a whitespace between selectors and variants
+		return errorf("%w", mf2.ErrSyntax)
```
janishorsts, line 657:
```suggestion
		return errorf("missing whitespace between selectors and variants: %w", mf2.ErrSyntax)
```

parse/parse.go
```diff
@@ -636,6 +636,11 @@ func (p *parser) parseMatcher() (Matcher, error) {
 
 selectorsLoop:
 	for {
+		itm := p.next()
+		if itm.typ != itemWhitespace {
+			return errorf("%w", unexpectedErr(itm, itemWhitespace))
+		}
+
```
janishorsts, line 643: when lexer lexes whitespace, it concatenates all whitespace characters and emits whitespace item. Effectively, the lexer won't emit two sequential whitespace items.  Please use `next()` instead of `nextNonWS()`!

parse/parse.go
```diff
@@ -636,7 +636,12 @@ func (p *parser) parseMatcher() (Matcher, error) {
 
 selectorsLoop:
 	for {
-		switch itm := p.nextNonWS(); itm.typ {
+		itm := p.next()
+		if itm.typ != itemWhitespace {
+			return errorf("%w", unexpectedErr(itm, itemWhitespace))
```
janishorsts, line 641:
```suggestion
			return errorf("missing whitespace before selector: %w", unexpectedErr(itm, itemWhitespace))
```

PR #241, opened 2024-09-18
Move data model errors to parser

mvilks:


parse/parse.go
```diff
@@ -648,6 +674,10 @@ selectorsLoop:
 		case itemEOF:
 			return errorf("%w", unexpectedErr(itm))
 		case itemVariable:
+			if _, ok := p.annotations[itm.val]; !ok {
```
janishorsts, line 677: Would it be better to compute direct/indirect annotations in the parseMatcher function instead? Otherwise, it computes annotations even when the message does not contain a matcher.

parse/parse.go
```diff
@@ -625,7 +638,35 @@ func isFallback(keys []VariantKey) bool {
...
+
+			return t.Annotation != nil
+		case LocalDeclaration:
+			if t.Variable != variable {
+				continue
+			}
+
+			if t.Expression.Annotation != nil {
+				return true
+			}
```
janishorsts, line 657: We require an annotation with a function. Annotation has optional function.

> Missing Selector Annotation:
> Each selector MUST be a variable that directly or indirectly references a declaration with a function.

janishorsts: I looked at the code and annotation is a function. MF2 does not use annotation anymore.

Rename to Function in another PR?!

Ignore me for now!

mvilks: Annotation is the function. The name remains from previous iterations where annotation was also a private or reserved stuff.

janishorsts (approved):
LGTM

PR #243, opened 2024-09-19
variable assignment is not allowed in attribute

janishorsts:


janishorsts:
Fixes unicode-org/message-format-wg#893

mvilks (approved):

mvilks:
Update the git submodule, too

PR #259, opened 2024-09-24
Save last called function and options in ResolvedValue

mvilks:


template/registry_testfunc.go
```diff
@@ -3,13 +3,16 @@ package template
...
+	"slices"
 	"strconv"
 
 	"go.expect.digital/mf2"
 	"golang.org/x/text/language"
 )
 
 // RegistryTestFunc is the implementation of the :test:function.
+//
+//nolint:gocognit
```
janishorsts, line 15: We should always attempt to resolve _gocognit_ issues.

template/registry_testfunc.go
```diff
@@ -25,6 +28,12 @@ func RegistryTestFunc(name string) func(*ResolvedValue, Options, language.Tag) (
 			return errorf("%w", mf2.ErrBadOperand)
 		}
 
+		// "If arg is the resolved value of an expression with a :test:function, :test:select, or :test:format
+		// annotation for which resolution has succeeded, then [..]" merge all options from resolved value.
+		if slices.Contains([]string{":test:function", ":test:format", ":test:select"}, operand.function) {
+			options = MergeOptions(operand.options, options)
```
janishorsts, line 34: I wonder if this should be the responsibility of `Template`. The primary benefit is that the registry function is slimmer. And adding a new registry function is easier.

```suggestion
			options = maps.Copy(operand.options, options)
```

janishorsts: Also, it enforces common behaviour across all registry functions.

template/registry_testfunc.go
```diff
@@ -3,13 +3,16 @@ package template
...
 
 	"go.expect.digital/mf2"
 	"golang.org/x/text/language"
 )
 
 // RegistryTestFunc is the implementation of the :test:function.
+//
+//nolint:gocognit
 func RegistryTestFunc(name string) func(*ResolvedValue, Options, language.Tag) (*ResolvedValue, error) {
 	if name != "format" && name != "select" {
```
janishorsts, line 17: Validating "function" is missing here

```suggestion
	if name != "format" && name != "select" && name != "function" {
```

template/registry_testfunc.go
mvilks: > Each function specification MAY have its own rules to preserve some options in the returned structure and discard others. In instances where a function specification does not determine whether an option is preserved or discarded, each function implementation of that specification MAY have its own rules to preserve some options in the returned structure and discard others.

So not a strong requirement, up to us what we do to options across different expressions.

template/registry_testfunc.go
mvilks: :test:function not used anywhere

template/template.go
```diff
@@ -103,6 +105,15 @@ func WithSelectKey(selectKey func(keys []string) string) ResolvedValueOpt {
 	}
 }
 
+// WithFunction applies a function name and options with which that function was called to the
+// [ResolvedValue]. It allows consecutive functions of the same type to preserve and re-use options.
+func WithFunction(funcName string, opts Options) ResolvedValueOpt {
```
janishorsts, line 110: Let's not expose this function, it has no practical use to others who want to register a new function.

```suggestion
func withFunction(funcName string, opts Options) ResolvedValueOpt {
```

